PR #11, opened 2015-12-16
use spaces everywhere

robertkowalski:


terinjokes:
There was a time in my life where I thought tabs were the best. Adjust
everything to 3 spaces if you want!

Except no one else has their editor configured correctly. :(

:+1:

PR #12, opened 2015-12-16
first console.log specs

robertkowalski:
adds first console.log specs.

I left some comments in the diff

maybe you can also take a look @domenic?


index.bs
```diff
@@ -87,9 +87,17 @@ The side effect of running the logger function with data and logLevel set to err
 
 The side effect of running the logger function with data and logLevel set to info.
 
-<h4><dfn method for="Console">log()</dfn></h4>
+<h4><dfn method for="Console">log([arg1], [arg2], ...)</dfn></h4>
 
-The side effect of running the logger function with data and logLevel set to log.
+Prints to the console as side effect of running the logger function if the logLevel is set to log. If the log function is called with no arguments, no output is written to the console. Each log entry from console.log gets an own line in the console. It is possible to pass format specifiers to the function in the first argument.
```
terinjokes: Should this be "and the logLevel is set to log."?

Would this algorithm be better with the logger function?

domenic:
This is exciting!

So, the setup that I think will work best is to leave the text for console.log() and friends the same as it is now:

> The side effect of running the logger function with data and logLevel set to log.

Maybe that could be clarified to

> Perform the side effects given by Logger(data, "log").

The key is then to move the algorithm you've worked on here into the header titled "Logger Function". (Which, if we do the above clarification, you'd probably want to add some text like `The logger function <dfn>Logger</dfn>(<var>data</var>, <var>logLevel</var>) is defined as follows:`)

This would allow you to not only specify log(), but also warn(), debug(), info(), etc.

index.bs
```diff
@@ -87,9 +87,17 @@ The side effect of running the logger function with data and logLevel set to err
...
+
+<emu-alg>
+1. If no argument given, _log_ prints nothing to the console
```
domenic: End sentence with a period. Also probably include "Abort these steps." Then the next step can be moved to the end, so that first you interpolate the string, then you print it.

index.bs
```diff
@@ -87,9 +87,17 @@ The side effect of running the logger function with data and logLevel set to err
...
+1. If at least one argument given, print it to the console on a new line
+2. If _arg1_ contains one or more format specifier, interpolate the string. Apply a type conversion to each argument following _arg1_ which matches the corresponding format specifier from left to right
```
domenic: This is OK, but could be more detailed (perhaps in a future commit). You'd ideally give an algorithm for converting a format string + subsequent args into a formatted string. (It might even help to write such an algorithm in JS and then transcribe it into prose; I do that a lot.)

If you do this, the next two steps will simply become part of the algorithm, and subtle cases like `%%` or `%%%` or `%!` or so on will become clear, and you'll also get an explicit listing of the allowed format specifiers and their corresponding type conversions.

robertkowalski:
thanks for all the feedback! 

i built a small Logger function with format specifiers in 9374ec4 and will work on the log algorithm in prose soon while trying to incorporate all your feedback

index.bs
robertkowalski: updated the PR. :)

robertkowalski:
hey @domenic hey @terinjokes i added a lot of specification for the logger function including the format specifiers. I also added an example for a nodejs implementation of the logger.

i need some feedback which areas need further improvements as I would like to merge the PR soon in order to continue on other parts of the spec

index.bs
```diff
@@ -24,9 +46,66 @@ Please join us in the <a href="https://github.com/terinjokes/console-spec/issues
...
+Each call of the Logger function prints a line which is ended by a newline.
+If the Logger function is called with no arguments, no output is written to the console.
+If the printing function gets multiple elements, it prints them separated by a space. It is possible to pass format specifiers to the function as the element of the arguments that should get printed.
+
+The Logger(logLevel, multipleArgs) function must run these steps:
+
+<emu-alg>
+  1. Let _firstElement_ be the first element of _multipleArgs_.
+  1. Let _rest_ be all elements following _firstElement_ in _multipleArgs_.
+  1. Let _formatter_ be a recursive function.
```
domenic: What is "a recursive function"?

index.bs
```diff
@@ -24,9 +46,66 @@ Please join us in the <a href="https://github.com/terinjokes/console-spec/issues
...
+  1. If no _firstElement_ given, do not print any output. Abort these steps.
+  1. If just _firstElement_ given, print _firstElement_ to the console. Abort these steps.
+  1. If _firstElement_ does not contain any format specifier, print _multipleArgs_ separated by one space to the console. Abort these steps.
+  1. Call _formatter_ with _multipleArgs_:
+    1. Let _target_ be the first element of _multipleArgs_.
```
domenic: Why are these steps nested under "Call formatter with multipleArgs"?

index.bs
```diff
@@ -24,9 +46,66 @@ Please join us in the <a href="https://github.com/terinjokes/console-spec/issues
...
+    1. Find the first possible format specifier from the left to the right in _target_.
+      1. If it is `%s`, let _converted_ be the result of new String(_current_).
+      1. If it is `%d` or `%i`, let _converted_ be the result of Number.parseInt(_current_, 10).
+      1. If it is `%f`, let _converted_ be the result of  Number.parseFloat(_current_, 10).
+      1. Replace _current_ in _target_ with _converted_.
+      1. Create _result_ which contains _target_ together with the elements of _multipleArgs_, but with _current_ removed.
+    1. If _target_ does not have any format specifiers left, return _result_ and end the recursive function. Print the result. Abort the steps.
+    1. If _result_ is just one element, return _result_ and end the recursive function. Print the result. Abort the steps.
+    1. Call _formatter_ with _result_ until it returns.
+</emu-alg>
```
domenic: I think maybe I get it. What you want is to create a second abstract operation, Formatter(multipleArgs), which is defined in a separate section similar to Logger. That will be better than nesting the definition of the formatter function inside the definition of Logger.

index.bs
```diff
@@ -24,9 +46,66 @@ Please join us in the <a href="https://github.com/terinjokes/console-spec/issues
...
+  1. Let _rest_ be all elements following _firstElement_ in _multipleArgs_.
+  1. Let _formatter_ be a recursive function.
+  1. If no _firstElement_ given, do not print any output. Abort these steps.
+  1. If just _firstElement_ given, print _firstElement_ to the console. Abort these steps.
+  1. If _firstElement_ does not contain any format specifier, print _multipleArgs_ separated by one space to the console. Abort these steps.
+  1. Call _formatter_ with _multipleArgs_:
+    1. Let _target_ be the first element of _multipleArgs_.
+    1. Let _current_ be the second element of _multipleArgs_.
+    1. Find the first possible format specifier from the left to the right in _target_.
+      1. If it is `%s`, let _converted_ be the result of new String(_current_).
```
domenic: do `ToString(_current_)` instead. `new String(_current_)` implies that by modifying `global.String` I can modify this result.

Makefile
```diff
@@ -0,0 +1,3 @@
+api:
+	curl https://api.csswg.org/bikeshed/ -F file=@index.bs | node_modules/.bin/emu-algify \
```
terinjokes: we probably need to include the `emu-algify` modules then.

index.bs
```diff
@@ -24,9 +46,66 @@ Please join us in the <a href="https://github.com/terinjokes/console-spec/issues
...
+      1. If it is `%d` or `%i`, let _converted_ be the result of Number.parseInt(_current_, 10).
```
domenic: For similar reasons, do `%parseInt%(_current_, 10)` and `%parseFloat%(_current_, 10)`. The ES spec defines the % versions to be the "original" versions of the functions, unaffected by changes to `global.Number`.

index.bs
```diff
@@ -128,7 +207,65 @@ Then, the side effect of running the logger function with arguments label and du
...
+      <th>Function</th>
+      <th>Type Conversion</th>
+    </tr>
+  </thead>
+  <tr>
+  <tr>
+    <td>%s</td>
+    <td>Element which substitutes is converted to a String</<td>
+    <td>new String(element)</td>
+  </tr>
```
domenic: Update these as mentioned above

domenic:
This is looking really good to me. The algorithm definition for the logger function is a bit confusing, but once we get that fixed up, IMO we should merge this and then consider moving this spec to be an official WHATWG spec, since it has enough value at that point to be worthwhile.

index.bs
terinjokes: Separating the formatting logic was part of my initial plan.

package.json
```diff
@@ -0,0 +1,25 @@
+{
```
domenic: Nits: directories is unnecessary; repository can be replaced with `"repository": "terinjokes/console-spec"`; bugs and homepage can be omitted.

promise-debounce is also not necessary for your workflow.

index.bs
```diff
@@ -24,9 +46,66 @@ Please join us in the <a href="https://github.com/terinjokes/console-spec/issues
...
+<div class="example">
+  The simplest way to implement the side effects of the Logger Function on the Node.js platform is to join the previously formatted arguments separated by a space and write the output to `stdout` or `stderr`.
+
+  Example implementation in Node.js:
+
+  <pre><code class="lang-javascript">
+    const util = require('util');
+    function print(logLevel, ...args) {
+
+      const message = util.format.apply(this, args);
```
domenic: Would this work as just `util.format(...args)`?

robertkowalski:
happy new year!

i incorporated the feedback, thank you! :)

domenic:
OK, this looks really good to me. @terinjokes, do you want to squash and merge and then transfer this repo to whatwg? I'll make you and @robertkowalski owners, and I have a follow-up PR in mind to clean up some minor wording things, but this is enough of a start I think we can call it a new WHATWG spec.

terinjokes:
@domenic doing a pass now, and I don't see anything worth commenting on, I'll happily squash and merge.

robertkowalski:
:heart_eyes:

PR #17, opened 2016-01-06
Tweaks for the move to WHATWG

domenic:
- Add deploy script
- Use resources.whatwg.org logo
- Update README
- Add @robertkowalski as an editor
- Tweak the header contents to align with most recent WHATWG convention

Depends on https://github.com/whatwg/resources.whatwg.org/pull/12


domenic:
I could use some help figuring out how to make a better makefile here (e.g. probably there should be a clean task and stuff); a good makefile could simplify the deploy script.

terinjokes:
@domenic I can clean up the Makefile this evening.

domenic:
ping

terinjokes:
I asked @robertkowalski to clean up the Makefile, since I've been flooded with other tasks this week.

domenic:
OK, let's see if Travis likes us when running for real...

domenic:
Nope; it can't figure out that we're on master... will push a few commits to try to figure it out.

robertkowalski:
yes i am also lookign right now

robertkowalski:
looks like you fixed it, but now it prompts to accept the key

robertkowalski:
maybe it gets fixed by `-o "StrictHostKeyChecking no"` -- did it work before?

domenic:
Yeah that's what I did in streams but I thought maybe it was because of some other streams stuff. I think the solution is https://docs.travis-ci.com/user/ssh-known-hosts/. Force-pushing over the last commit with that; we'll see.

domenic:
Grr worked through a few fixes but now it's asking for a password. OK, going to try https://docs.travis-ci.com/user/deployment/custom/ instead of using `scp -i`

terinjokes:
@domenic why not try using? https://docs.travis-ci.com/user/deployment/script/

domenic:
@terinjokes I've never used that before; how would it help?

I think the problem is that I forgot to nest the secure env variable under env: global: in .travis.yml... we'll see if this next build works...

domenic:
Success, sort of!! Not sure what's up with that first line lol https://console.spec.whatwg.org/

PR #19, opened 2016-01-08
use a makefile

robertkowalski:
hi @domenic,

i wasn't sure how much you wanted to be part of the makefile. I converted everything, but maybe you want to separate the upload and deploy. i also tried to make it generic, so it should work for most other specs by just changing the spec variable at the top.

i added a lot of comments to explain what each section does. I learned most about make from isaacs, this video is great: https://www.youtube.com/watch?v=dsqBSgdQz_8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,54 @@
-api:
-	mkdir -p console.spec.whatwg.org && \
-	curl https://api.csswg.org/bikeshed/ -F file=@index.bs | node_modules/.bin/emu-algify \
-	--throwing-indicators > console.spec.whatwg.org/index.html
+spec = console
+targetfolder = $(spec).spec.whatwg.org
+server = $(spec).spec.whatwg.org
+
+branch = $(shell git rev-parse --abbrev-ref HEAD)
+
+# find all bikeshed files and create a target for them
+# ./index.bs gets console.spec.whatwg.org/index.html
+markupfiles := $(shell find . -name '*.bs' \
+								| sed 's|.bs|.html|g' \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,4 +1,52 @@
...
+	@echo "Living standard output to $(targetfolder)"
+	@echo ""
+	@find $(targetfolder) -print
+	@echo ""
+endif
+
+# upload to server
+upload:
+	@openssl aes-256-cbc -K $(encrypted_b9b018a1d67d_key) -iv $(encrypted_b9b018a1d67d_iv) -in console_spec_id_rsa.enc -out console_spec_id_rsa -d
+	@scp -r -i console_spec_id_rsa $(targetfolder) $(DEPLOY_USER)@$(server):
```
robertkowalski: caution: i was not able to test this part

deploy.sh
```diff
@@ -1,31 +1,4 @@
 #!/bin/bash
```
terinjokes: Any reason we need to keep this file around?

robertkowalski: no - i just wasn't sure about how it is hooked in into other systems, so i left it there for now

robertkowalski: i also wasn't sure if we want to move completely to make, also for the deployment

domenic: We can move completely to make. Nice work! Switch over the .travis.yml I guess.

domenic:
Lovely; merged into #17 as https://github.com/whatwg/console/commit/624627ffec1f2c09a94c1da3e09821afcfd53dbe

PR #28, opened 2016-01-12
Fix #24: abstract operations should now be autolinked

### LINKED ISSUE #24: Abstract operations are not getting linked correctly
E.g. ToString should link to the ES spec, and Formatter() should link to the formatter AO.

This is probably a bug in emu-algify, so assigning to me.

domenic:
Still some remaining issues: https://github.com/bterlson/ecmarkup/issues/77 and https://github.com/bterlson/ecmarkup/issues/76 but they are not our fault.


terinjokes:
:+1:

PR #30, opened 2016-01-13
add difference when browsers return undefined

robertkowalski:
it looks like FF is working in an async way (returns undefined
before the logger prints) and that the Chrome implementation is
blocking until the printer printed.


terinjokes:
LGTM :+1:

robertkowalski:
merged as a869f345b77efc134b93864602243d1e340cf1d9

Issue #25, opened 2016-01-10
Clarify formatting specifiers a bit


domenic:
Right now the section https://console.spec.whatwg.org/#formatting-specifiers is kind of disconnected from the main text. https://console.spec.whatwg.org/#formatter handles some of them, but the ones that require complex UI are ignored by the rest of the spec's algorithms.

I am not sure how best to reconcile this. Maybe Formatter() should reference the tables in #formatting-specifiers, and remove the algorithm steps? Maybe Formatter() should get an additional line for each of %o, %O, or %c? But these don't act like type conversions, exactly...

In any case one goal should be to ensure there is only one normative source of truth---either the algorithm or the tables. If we decide on the algorithm being normative, then we can keep the tables, we just need to mark them non-normative.

terinjokes:
My vote would be to have the algorithms as normative and have the tables as non-normative reference.

robertkowalski:
i agree with terin, i would also prefer that and also mentioned it in https://github.com/whatwg/console/issues/6#issuecomment-170396383

Issue #6, opened 2014-10-08
Investigate optional feature support advertisement 


terinjokes:

robertkowalski:
If I interpret the title right I think the support for the format specifiers 

``` 
%o     Displays as expandable DOM element
%O    Displays as expandable JS Object
%c     Applies provided CSS
```

should be optional. The expandable elements are difficult to realise in a terminal, e.g. for platforms like Node.js. Basically the same applies to the CSS styling which is just partly possible (colors would be possible)

robertkowalski:
@terinjokes ping

robertkowalski:
did i understand intention of the title properly?

terinjokes:
Yes, I believe so.

I think the idea was that the JS would be able to see that expandable elements aren't possible and pick a different way to long, but maybe that's just too complicated.

domfarolino:
I feel like a lot of the text in https://console.spec.whatwg.org/#printer (specifically user experience innovation and common object formats) allows for optional feature support, which is referenced in the https://console.spec.whatwg.org/#formatting-specifiers table. I think I'll add a note in https://github.com/whatwg/console/issues/98 so that we know to mention in some capacity, that the application of styles are up to the implementation and then close this out.

PR #33, opened 2016-01-18
mark formatting-complex-ui section as non-normative

robertkowalski:


robertkowalski:
see: 

https://github.com/whatwg/console/issues/25 
https://github.com/whatwg/console/issues/6

terinjokes:
I'm not sure I agree with marking `%o` and `%O` as non-normative. Fine with marking the section, as long as we also include them in the algorithm.

domenic:
So I think the best plan would be:

- Add *optional* (not non-normative) steps to the algorithm for %o and %O.
- Mark the entire formatting specifiers section as non-normative. And maybe have just one section with a paragraph separating the two tables like "the following additional format specifiers require a complex UI, and as such are optional for implementations:".
- Move the formatting specifiers section into a subsection of Printer, maybe?

PR #37, opened 2016-01-27
Remove note about async execution, since behavior in Firefox matches spec as of Bug 1044365 (version 46)

bgrins:
I've updated the behavior in Firefox to match the spec in https://hg.mozilla.org/mozilla-central/rev/2903b0280572, so can the note be removed?  Version 46 will be promoted to the Dev Edition channel later this week.

### DIFF
--- a/NOTES.md
+++ b/NOTES.md
@@ -90,22 +90,3 @@ Chrome:
 console.log("%s %snewword", "duck")
 duck %snewword
 ```
-
-## Async printing execution
-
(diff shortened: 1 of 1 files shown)


robertkowalski:
great, thank you, LGTM

robertkowalski:
merged as c20dd0bcade15784ba2be3f7d650632aaced9926 with linebreak after max 72 chars in commit message

PR #38, opened 2016-01-28
Remove mention of Firefox 'nan' vs. 'NaN' inconsistency.

frewsxcv:


robertkowalski:
Thank you!

I merged https://github.com/whatwg/console/commit/2fde8b5f1c9ef534284e6cab625a86d181c2f5d7 yesterday which removes another difference from the list.

I would like to have it consistent so we should decide on removing it or adding a link to the issue etc.

@terinjokes @domenic @frewsxcv do you have any opinions what might work best?

domenic:
I like removing in general, but maybe some kind of appendix of "recently fixed bugs" or similar would be interesting. Alternately we could just rely on the Git history.

robertkowalski:
@frewsxcv is it OK for you to delete the entry and add a good commit message (maybe also mentioning in which FF version it was fixed and the link https://bugzilla.mozilla.org/show_bug.cgi?id=1239506) or is that completely impractical for you?

frewsxcv:
@robertkowalski Let me know how that looks

robertkowalski:
LGTM thank you! :+1:

robertkowalski:
merged as d9e224a

PR #40, opened 2016-01-28
Make format specifier spec more rigorous

### LINKED ISSUE #25: Clarify formatting specifiers a bit

### LINKED ISSUE #36: %o specifier uses type-specific formatter
(Forking this from discussion in #27)

Chrome's implementation right now is: 

* `%o`  - display using the user agent's default formatting for that object type. For example: `console.log('%o', obj)` is identical to `console.log(obj)`.
* `%O`  - display as a javascript object. For example: `console.log('%O', obj)` is identical to `console.dir(obj)`.

Currently, Chrome logs elements as a tree (dirxml-style), but we plan to change that very soon (to log them as objects (dir-style). As a result, `%o` will change because it just follows our default formatting. There is currently no format specifier to force dirxml-style logging of objects. (source: [`_formatWithSubstitutionString`](https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/devtools/front_end/console/ConsoleViewMessage.js&q=f:consoleviewmessage%20formatters.o&sq=package:chromium))

`%o` is a generalized specifier for all js objects. It will use whatever formatter we think is best. They can be one of the following:  `array, error, function, generator, iterator, map, node, object, set, string`. (source: [`_customFormatters `](https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/devtools/front_end/console/ConsoleViewMessage.js&q=f:consoleviewmessage%20%22:%20this._formatParameter%22&sq=package:chromium&type=cs&))

Here's the difference of `%o` and `%O` on an array:

image

domenic:
Closes #25. Closes #36 by incorporating @paulirish's feedback that %o is a bit more generic than "expandable DOM object". Closes #33 by superseding it.

Let me know what you guys think :)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -63,7 +63,7 @@ The logger operation accepts a log level and a List of other arguments. Its main
 
 <h3 id="formatter" aoid="Formatter" nothrow>Formatter(<var>args</var>)</h3>
 
-The formatter operation tries to format the first argument provided, using the other arguments. It will try to format the input until no <a>formatting specifiers</a> are left in the first argument, or no more arguments are left. It returns a List of objects suitable for printing.
+The formatter operation tries to format the first argument provided, using the other arguments. It will try to format the input until no formatting specifiers are left in the first argument, or no more arguments are left. It returns a List of objects suitable for printing.
 
 <emu-alg>
   1. Let _target_ be the first element of _args_.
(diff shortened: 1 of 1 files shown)


robertkowalski:
wow thank you!

i think it looks great and I read it three times and it all makes sense.

much better than `Formatting specifiers that require a complex UI` :wink: 

I have some more time since today as I just finished another project - looking forward to collaborate more again

terinjokes:
:+1:

terinjokes:
Merged as https://github.com/whatwg/console/commit/c27e994a82c37fb6729b3e21fde8447b3cafc6f6.

PR #41, opened 2016-01-28
console.assert improvements

### LINKED ISSUE #4: allow formatting specifiers in console.assert.
All the logging functions currently allow the use of formatting specifiers, with the exception of assert. To avoid confusion for web developers, the assert function should work with formatting specifiers.

```javascript
console.assert(false, 'string: %s', 'test');
//-> Assertion failed: string: test
```

robertkowalski:
- Adds found differences for Chrome & Firefox.
 - Fixes method signature to support multiple arguments and not just one message. 

As the logger function is called if the assertion fails, console.assert MUST support format specifiers, which closes #4

### DIFF
--- a/NOTES.md
+++ b/NOTES.md
@@ -78,3 +78,32 @@ Chrome:
 console.log("%s %snewword", "duck")
 duck %snewword
 ```
+
+## console.assert
+
+```
+FF / Edge:
+
+console.assert(false, "robert keeps %s on his balcony", "plaices")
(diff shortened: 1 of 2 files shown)


terinjokes:
LGTM :+1:

domenic:
LGTM, maybe add "closes #4" to the second commit.

robertkowalski:
merged with the edge addition - but i forgot the fixes comment :(

PR #42, opened 2016-01-29
Add note about how the logger should print before the return value is…

### LINKED ISSUE #39: Add a NOTE about "not async" behavior
#38 and the related Firefox fix could maybe be called out more explicitly in the spec. I would suggest a non-normative NOTE something like this, after the Logger algorithm.

> It's important here that the printing occurs before returning from the algorithm. Many developer consoles print the result of the last operation entered into them. In such consoles, when a developer enters `console.log("hello!")`, this should first print `"hello!"`, _then_ the `undefined` return value from the `console.log` call.

You could even include an example screenshot showing this behavior.

bgrins:
Fixes #39.  Let me know if it needs some changes.

* I'm not sure if the note is non-normative by default or if there's some extra markup needed for that.
* Also this adds an `images/`directory with a small screenshot of the expected behavior.  I can update the screenshot or put it elsewhere if needed.


index.bs
```diff
@@ -61,6 +61,12 @@ The logger operation accepts a log level and a List of other arguments. Its main
   1. Return *undefined*.
 </emu-alg>
 
+<div class="note">
+  It's important that the printing occurs before returning from the algorithm. Many developer consoles print the result of the last operation entered into them. In such consoles, when a developer enters <code>console.log("hello!")</code>, this should first print "hello!", then the undefined return value from the console.log call.
+
+  <img src="images/print-before-returning.png" />
```
domenic: No ` /`, and needs an alt describing the image in some detail.

domenic:
> I'm not sure if the note is non-normative by default or if there's some extra markup needed for that.

Non-normative by default; all good.

bgrins:
Updated the image markup / alt text, and also updated the Makefile to use `@cp` instead of `shell` to copy the images directory.

index.bs
```diff
@@ -61,6 +61,12 @@ The logger operation accepts a log level and a List of other arguments. Its main
   1. Return *undefined*.
 </emu-alg>
 
+<div class="note">
+  It's important that the printing occurs before returning from the algorithm. Many developer consoles print the result of the last operation entered into them. In such consoles, when a developer enters <code>console.log("hello!")</code>, this should first print "hello!", then the undefined return value from the console.log call.
+
+  <img src="images/print-before-returning.png" alt="Screenshot showing that output should be printed before the developer console prints a return value">
```
domenic: This is pretty good, but I might suggest something like "screenshot showing a developer console in which console.log("hello!") is entered, immediately after which "hello!" is printed, followed by the undefined return value"

bgrins:
Alright, updated the alt text to match suggestion

domenic:
This LGTM; will defer to the editors on the images/ and such. They might also want the commit message to follow https://github.com/erlang/otp/wiki/Writing-good-commit-messages or include "Fixes #39" but @robertkowalski has been fixing that kind of thing as he merges.

robertkowalski:
LGTM apart from the commit message - thank you @bgrins :+1: 

will set it to max 72 chars char length and add a "fixes #39" - next time you'll have to fix it on your own @bgrins :))

robertkowalski:
merged as e5f14f7b595197103fd72656a90e586758290522 with fixed commit message

thanks again, good work!

PR #44, opened 2016-02-08
Fix #2: `console` should always be visible

### LINKED ISSUE #2: `console` should always be visible
Add language that ensures that `console` is always visible and usable to scripts, even if the developer console has not been opened or does not exist.

robertkowalski:
- modify IDL, explicitly exposing console
 - add language that implementations should buffer
 - add additional note that console is always visible

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -210,8 +210,9 @@ partial interface WorkerGlobalScope {
 </pre>
 
 <p class="note">
-It is important that `console` is always visible and usable to scripts,
-even if the developer console has not been opened or does not exist.
+It is important that <code>console</code> is always visible
+and usable to scripts, even if the developer console has not been opened or
+does not exist.
 </p>
 
 <h3 id="logging">Logging methods</h3>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -139,6 +139,9 @@ The printer operation is implementation-defined. It accepts a log level indicati
 
 By the time the printer operation is called, all format specifiers will have been taken into account, and any arguments that are meant to be consumed by format specifiers will not be present in <var>args</var>. The implementation's job is simply to print the List.
 
+If the console is not open, implementations should choose to buffer messages
```
domenic: not open "when the printer operation is called".

Remove "choose".

Maybe add something like ", up to an implementation-chosen limit (typically on the order of at least 100)".

index.bs
```diff
@@ -205,6 +208,11 @@ partial interface WorkerGlobalScope {
 };
 </pre>
 
+<p class="note">
+It is important that `console` is always visible and usable to scripts,
```
domenic: I'm not sure backticks will work here; you might need to use `<code>`. Check the output.

index.bs
robertkowalski: > not open "when the printer operation is called".

great this more precise, <3 it

> Maybe add something like ", up to an implementation-chosen limit (typically on the order of at least 100)".
 
good call... wanted to write something about the limits but had a hard time finding the right language... fixing...

domenic:
LGTM!

robertkowalski:
merged as 1b525c6a149c3f6f36d3dbcb9716ed3f34c78f57

thanks for the good feedback :)

PR #58, opened 2016-05-04
add defaults to time, timeEnd, and assert

### LINKED ISSUE #54: console.time with no arguments
What should be the behavior of `console.time` with no arguments?

Currently the spec shows a single non-optional argument:

```
interface Console {
  void time(DOMString label);
  void timeEnd(DOMString label);
};
```

Current behavior of `console.time()` and `console.timeEnd()`:
* Firefox - silently does nothing
* Chrome - silently does nothing
* node.js - starts/stops an unnamed timer
* Safari - throws an exception, warning the user they need to provide a label
* Edge - starts/stops a timer named `"default"`, console.warns when called and the timer is/not running

Current behavior of `console.time("name")` and `console.timeEnd("name")`:
* All environments behave the same

I have to admit I rather like both node and Edge's behavior here. WebKit's is nice in that it indicates a likely user error that would otherwise cause a developer to scratch their head, but an exception is overkill. node and Edge make it simple to do a quick and dirty time & timeEnd, which is what developers will want most of the time.

I would suggest a move in the Edge direction.

Proposed:

```
interface Console {
};
```

Playing around with Edge's implementation, it seems that any falsey value (e.g. `false`, `0`, `NaN`, `""`, ...) is coerced to the string `"default"`. I don't know of an IDL attribute for this kind of behavior (`[TreatFalseyAs="default"]`), or if it would be desirable to spec this quirk. However, it does seem very user friendly.

Comparing node and Edge, node is rather ambiguous. node seems to treat `label` as a value and ambiguously displays string names for values that aren't treated as strings internally. For instance `console.time(123)` and `console.time('123')` are treated uniquely, but based on warning messages from node they are not distinguishable. Likewise `undefined` versus `"undefined"`. I could see this being extended to with object / Symbol values.

Ultimately I don't think non-string values as identifiers are necessary for `console.time()`. And given all browsers currently expect strings, it seems easier to stick with that expectation.

_Update: Added node.js to the discussion._

terinjokes:
Resolves #54

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -196,8 +196,8 @@ interface Console {
   void groupEnd();
 
   // Timing
-  void time(DOMString label);
-  void timeEnd(DOMString label);
+  void time(optional DOMString label = "default");
+  void timeEnd(optional DOMString label = "default");
 };
 
 partial interface Window {
--- a/index.bs
+++ b/index.bs
@@ -179,7 +179,7 @@ implementation-chosen limit (typically on the order of at least 100).
 [NoInterfaceObject, Exposed=(Window,Worker)]
 interface Console {
   // Logging
-  void assert(boolean condition, any... data);
+  void assert(optional boolean condition = false, any... data);
   void clear();
   void count(optional DOMString label = "");
   void debug(any... data);


domenic:
LGTM

PR #61, opened 2016-06-11
document differences for console.count argument handling

robertkowalski:


robertkowalski:
ping @domenic @terinjokes

PR #83, opened 2017-01-14
Meta: add commit and branch snapshots

domenic:
This moves the deploy process to be in line with most other WHATWG specifications; see https://github.com/whatwg/meta/issues/2. It also adds the missing dfn.js file to get popups when clicking on <dfn> elements and makes other small tweaks.

I know @robertkowalski did a lot of work to move the deploy process into the makefile, but we're pretty much using this similar deploy.sh script for all other repos, so it's probably best to be uniform.

Let's see if Travis likes this as a PR...


domenic:
Tada:

- https://console.spec.whatwg.org/branch-snapshots/snapshots/
- https://console.spec.whatwg.org/commit-snapshots/e51efb2f3c8bbdd9bc6f71cb9f02ab9999a493c6/

annevk:
I guess at this point we should make this use the central deploy script?

domenic:
We'd need to add emu-algify to the central deploy script first, for both this and streams... I guess I should work on that.

robertkowalski:
wow thats super cool!

(and no worries, i don't cling to to the makefile)

> I guess at this point we should make this use the central deploy script?

I assume this isn't blocking the merge?

robertkowalski:
+1

PR #86, opened 2017-02-25
console.assert should always call logger with non-empty data (fixes #85)

### LINKED ISSUE #85: Clarify that console.assert should log even without formatting string
https://console.spec.whatwg.org/commit-snapshots/36c3ee56b8471c82ec5ea88dcbaf223235b22a3e/#assert

> If condition is false, perform Logger\("error", data\)\.

A call like `console.assert(false)` would then call `Logger("error", [])`, but the implementation of `Logger` states as the first item:

> 1. If `args` is empty, abort these steps.

I would expect the behavior to log something though - and the participants in the discussion around #56 seem to make the same assumption.

When calling `console.assert(false)` in various browsers, something is logged in that case:

- Chrome prints `Assertion failed: console.assert` with level *error*
- Safari prints `Assertion Failed` with level *error*
- Firefox prints an empty line with level *error*
- Edge prints `undefined` with level *error*
- Node throws an `AssertionError: false == true` (`false` is static, same result for `console.assert(0)`)

I'd propose to add a section that clarifies this:

1. If condition is *true*, abort these steps.
2. If *data* is not empty, perform Logger("error", *data*). Abort these steps.
3. Let *message* be the string "Assertion failed".
4. Perform Logger("error", [*message*])

Should we make a clear recommendation of what the String in (3) should look like? Or leave it open to implementers, e.g. just say "a string, indicating that some assertion failed (e.g. ...)"?

---

For reference, calling `console.assert(false, "foo")` in various browsers:

- Chrome prints `Assertion failed: foo` with level *error*
- Safari prints `Assertion Failed: foo` with level *error*
- Firefox prints `foo` with level *error*
- Edge prints `foo` with level *error*
- Node throws an `AssertionError: foo`

EDIT: Didn't have access to Edge, added @domenic's findings.

sirlantis:
The specification says that  Logger should not perform anything
if data is empty. However, console.assert() should not be required
to provide an argument besides the condition to log something,
which matches what most implementations do.

This clarifies that section in the spec by providing an algorithm
as well as a recommendation of prefixing messages without destroying
formatting capabilities.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -231,7 +231,7 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
   1. If _data_ is empty, append _message_ to _data_.
   1. Otherwise, implementations should perform these substeps:
     1. Let _first_ be the first element of _data_.
-    1. If _first_ is not a string, prepend _message_ to _data_. Abort these substeps.
+    1. If Type(_first_) is not String, prepend _message_ to _data_. Abort these substeps.
     1. Let _concat_ be the concatenation of _message_, U+003A COLON (:), U+0020 (SPACE), and _first_.
     1. Set _data[0]_ to _concat_.
   1. Perform Logger("error", _data_).
--- a/index.bs
+++ b/index.bs
@@ -232,7 +232,7 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
   1. Otherwise, implementations should perform these substeps:
     1. Let _first_ be the first element of _data_.
     1. If Type(_first_) is not String, prepend _message_ to _data_. Abort these substeps.
-    1. Let _concat_ be the concatenation of _message_, U+003A COLON (:), U+0020 (SPACE), and _first_.
+    1. Let _concat_ be the concatenation of _message_, U+003A COLON (:), U+0020 SPACE, and _first_.
     1. Set _data[0]_ to _concat_.
   1. Perform Logger("error", _data_).
 </emu-alg>
--- a/index.bs
+++ b/index.bs
@@ -223,7 +223,7 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
 
 <h3 id="logging">Logging methods</h3>
 
-<h4 id="assert" oldids="assert-condition-data,dom-console-assert" method for="console">assert(<var>condition</var>, ...<var>data</var>)</h4>
+<h4 id="assert" oldids="assert-condition-data,dom-console-assert" method for="console">assert(<var ignore=''>condition</var>, ...<var>data</var>)</h4>
 
 <emu-alg>
   1. If _condition_ is true, abort these steps.
--- a/index.bs
+++ b/index.bs
@@ -223,7 +223,7 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
 
 <h3 id="logging">Logging methods</h3>
 
-<h4 id="assert" oldids="assert-condition-data,dom-console-assert" method for="console">assert(<var ignore=''>condition</var>, ...<var>data</var>)</h4>
+<h4 id="assert" oldids="assert-condition-data,dom-console-assert" method for="console">assert(<var>condition</var>, ...<var>data</var>)</h4>
 
 <emu-alg>
   1. If _condition_ is true, abort these steps.
@@ -231,7 +231,7 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
   1. If _data_ is empty, append _message_ to _data_.
   1. Otherwise, implementations should perform these substeps:
     1. Let _first_ be the first element of _data_.
-    1. If _first_ is not a string, prepend _data_ with _message_. Abort these substeps.
+    1. If _first_ is not a string, prepend _message_ to data_. Abort these substeps.
     1. Let _concat_ be the concatenation of _message_, U+003A COLON (:), U+0020 (SPACE), and _first_.
     1. Replace the first element of _data_ with _concat_.
   1. Perform Logger("error", _data_).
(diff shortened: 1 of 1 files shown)


sirlantis:
In the issue, I have a if/else regarding "first element is a string" while here I have guarding if. Let me know what your preference is - from other specs it seems "abort these substeps" is quite common to reduce nesting.

index.bs
```diff
@@ -223,9 +223,19 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
 
 <h3 id="logging">Logging methods</h3>
 
-<h4 id="assert" oldids="assert-condition-data,dom-console-assert" method for="console">assert(<var>condition</var>, ...<var>data</var>)</h4>
+<h4 id="assert" oldids="assert-condition-data,dom-console-assert" method for="console">assert(<var ignore=''>condition</var>, ...<var>data</var>)</h4>
```
sirlantis: Trying to avoid a warning here:

```
WARNING: The following <var>s were only used once in the document:
  'condition'
If these are not typos, please add an ignore='' attribute to the <var>.
```

index.bs
```diff
@@ -225,7 +225,17 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
 
 <h4 id="assert" oldids="assert-condition-data,dom-console-assert" method for="console">assert(<var>condition</var>, ...<var>data</var>)</h4>
 
-If <var>condition</var> is false, perform Logger("error", <var>data</var>).
+<emu-alg>
+  1. If _condition_ is true, abort these steps.
+  1. Let _message_ be a string without any formatting specifiers indicating generically an assertion failure (such as "Assertion failed").
+  1. If _data_ is empty, append _message_ to _data_.
+  1. Otherwise, implementations should perform these substeps:
```
domenic: remove "implementations should"

index.bs
```diff
@@ -225,7 +225,17 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
...
+    1. Let _first_ be the first element of _data_.
+    1. If _first_ is not a string, prepend _message_ to _data_. Abort these substeps.
+    1. Let _concat_ be the concatenation of _message_, U+003A COLON (:), U+0020 (SPACE), and _first_.
+    1. Replace the first element of _data_ with _concat_.
```
domenic: This can use the syntax "Set _data_[0] to _concat", optionally.

index.bs
```diff
@@ -225,7 +225,17 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
...
 <h4 id="assert" oldids="assert-condition-data,dom-console-assert" method for="console">assert(<var>condition</var>, ...<var>data</var>)</h4>
 
-If <var>condition</var> is false, perform Logger("error", <var>data</var>).
+<emu-alg>
+  1. If _condition_ is true, abort these steps.
+  1. Let _message_ be a string without any formatting specifiers indicating generically an assertion failure (such as "Assertion failed").
+  1. If _data_ is empty, append _message_ to _data_.
+  1. Otherwise, implementations should perform these substeps:
+    1. Let _first_ be the first element of _data_.
+    1. If _first_ is not a string, prepend _message_ to _data_. Abort these substeps.
```
domenic: "is not a string" is kind of ambiguous. What do browsers do if it is `{ toString() { "some message" } }`? Do they convert it to a string?

If we want a type test here, this should be "If Type(_first_) is String, ...". But I am not sure that's what we want...

index.bs
domenic: Oh, I see, this was meant to be a "should" specifically. In that case this is fine, nevermind!

index.bs
sirlantis: The intention was a type test, mainly because `concat` is not well-defined for non-string types. An alternative would have been to do the following in the next line:

> Let _concat_ be the concatenation of _message_, U+003A COLON (:), U+0020 (SPACE), and [toString](https://tc39.github.io/ecma262/#sec-tostring)(_first_).

However that would lead to the following behavior, which is unintended and also would not match any current behavior:

```
console.assert(false, { foo: bar }) // "Assertion failed: [Object object]"
```

Will push a commit with the proper type test notation.

sirlantis: Sadly can't phrase it that easily. Only correct statement is: No implementation that concatenates prints `[Object object]`. Only Safari and Chrome have a prefix (and therefore a desire to concat:

- Safari concats and does a type test.
- Chrome doesn't concat (and therefore breaks formatting capabilities).

---

Test cases:
- `console.assert(false, { answer: 42 })`
- `console.assert(false, "foo: %s", "bar")`

Results:

- Chrome
    - `Assertion failed: Object {answer: 42}` (Object is expandable)
    - `Assertion failed: foo: %s bar`
- Safari:
    - `Assertion Failed – {answer: 42}` (Object behind long-dash which is their "rest" divider)
    - `Assertion Failed: foo: bar`
- Firefox:
    - `Object { answer: 42 }` (Object is expandable)
    - `foo: bar`
- Edge:
    - `[Object object]`
        - Note: Edge's `Printer` is a simple `toString` and does not print objects nicely - even for `console.log`
    - `foo: bar`

domenic: Fair enough!

index.bs
```diff
@@ -225,7 +225,17 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
...
 
-If <var>condition</var> is false, perform Logger("error", <var>data</var>).
+<emu-alg>
+  1. If _condition_ is true, abort these steps.
+  1. Let _message_ be a string without any formatting specifiers indicating generically an assertion failure (such as "Assertion failed").
+  1. If _data_ is empty, append _message_ to _data_.
+  1. Otherwise, implementations should perform these substeps:
+    1. Let _first_ be the first element of _data_.
+    1. If Type(_first_) is not String, prepend _message_ to _data_. Abort these substeps.
+    1. Let _concat_ be the concatenation of _message_, U+003A COLON (:), U+0020 (SPACE), and _first_.
```
domenic: Final nit: I think this should be "U+0020 SPACE", no parens.

sirlantis: Will change it. JFYI though, this is copied verbatim from the https://console.spec.whatwg.org/#count section:

> Let concat be the concatenation of label, U+003A COLON (:), U+0020 (SPACE), and ToString(called).

domenic:
@domfarolino feel free to merge PRs!

PR #87, opened 2017-02-27
Remove unorthodox parens around symbol name SPACE

sirlantis:
As mentioned in #86 so that the next person copy-pasting doesn't run into this.

Seems to be more consistent with other WHATWG specs, like https://html.spec.whatwg.org/

> [...] A U+0054 LATIN CAPITAL LETTER T character (T) or a U+0020 SPACE character [...]
> [...] of the string "CACHE", a single U+0020 SPACE character, the string "MANIFEST" [...]

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -235,7 +235,7 @@ If possible for the environment, clear the console. Otherwise, do nothing.
 
 <emu-alg>
   1. Let _called_ be the number of times _count_ has been invoked (including this invocation) with the provided _label_.
-  1. Let _concat_ be the concatenation of _label_, U+003A COLON (:), U+0020 (SPACE), and ToString(_called_).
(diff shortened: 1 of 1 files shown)


sirlantis:
(very low-impact quality PR - just since I was touching that file anyway)

domfarolino:
I don't know if there's some standard around how unicode characters should be communicated in a spec but I like how [Web IDL](heycam.github.io/webidl/) handles it. Just `CMD+F` in the spec looking for `U+` and you'll see examples like: `U+005F LOW LINE ("_")` where the "format" is `CODE NAME ("char_here")` which would give `U+003A COLON (":"), U+0020 SPACE (" ")` in [`2.1.3. count(label)`](https://console.spec.whatwg.org/#count).

Feel free ignore this comment, as it is a low-impact comment on what you said is a low-impact PR 😉

domenic:
The standard is https://infra.spec.whatwg.org/#code-points :)

Issue #88, opened 2017-03-02
Validate console.count IDL against implementations


domfarolino:
This is an extension of #26 where @domenic and I discussed implementation differences regarding `console.time/timeEnd`.

----
## [count](https://console.spec.whatwg.org/#count) ([IDL](https://console.spec.whatwg.org/#idl-index))

With count there are two things to test (similarly to time):

1. **The implementation of the IDL's default parameter `label = ""`**
   - This involves ensuring that `console.count()` and `console.count(undefined)` not only work, but also produce identical results
2. **ES [`ToString()`](https://tc39.github.io/ecma262/#sec-tostring) conversion is being delegated to**
   - This involves ensuring that `label.toString()` gets called if the count `label` is an object.

## Tests

- Test for (1) is in this [bin](https://jsbin.com/rituqot/edit?js,console)
- Test for (2) is in this [bin](https://jsbin.com/kibefuj/edit?js,console)

## Results:

- **Firefox**:
    - Fails (1) in that `count()` behaves differently than `count(undefined)`.
        - :x: `count()` uses label "no label" instead of `"default"`
        - :x: `count(undefined)` uses label `"undefined"` instead of `"default"`
    - Fails in that errors are never propagated to the user (2)
        - :white_check_mark: Calls `label.toString()` when label is object.
        - :x: Does not re-throw conversion error to user AND continues with `count()` behavior
- **Chrome + Canary**:
    - Fails (1) in that `count()` behaves differently than `count(undefined)`.
        - :x: `count()` uses label `""` instead of `"default"`
        - :x: `count(undefined)` uses label `"undefined"` instead of `"default"`
    - Fails both parts of (2)
        - :x: Does not call `label.toString()` ever (just like `time`)
        - :x: Obviously never re-throws errors since custom conversion never takes place
        - This actually is good, because fixing `time` may incidentally fix `count`!
- **Edge**:
    - Passes (1).
        - :x: `count()` uses label `""` instead of `"default"`
        - :white_check_mark:
    - Fails in that errors are never propagated to the user (2)
        - :white_check_mark: Calls `label.toString()` when label is object.
        - :x: Does not re-throw conversion error to user AND continues with `count()` behavior
- **Safari + Tech Preview (24)**:
    - Fails (1) in that `count()` behaves differently than `count(undefined)`
        - :x: `count()` uses label "Global" instead of `"default"`
        - :x: `count(undefined)` uses label `"undefined"` instead of `"default"`
[...]

PR #91, opened 2017-03-07
Modify count() IDL default param

domfarolino:
As suggested in #88

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -187,7 +187,7 @@ namespace console { // but see namespace object requirements below
   // Logging
   void assert(optional boolean condition = false, any... data);
   void clear();
-  void count(optional DOMString label = "");
+  void count(optional DOMString label = "default");
   void debug(any... data);
   void error(any... data);
   void info(any... data);


robertkowalski:
> same for count btw

but depends on implementors buy-in, maybe with the follow up issues easier to solve, things get easily lost over time

domfarolino:
> same for count btw

Sorry I'm a little confused, since this is modifying the `count` IDL. Do you think we should hold off on merging, or at least open some issues before we merge?

JosephPecoraro:
I still think having a `console.count` for a unique place in code is a useful concept, and it would be most natural as the no-argument version. However, I don't seem to have convinced anyone of my position though, heh.

robertkowalski:
> Sorry I'm a little confused, since this is modifying the count IDL. Do you think we should hold off on merging, or at least open some issues before we merge?

ah no sorry, i was just tired and messed things up 

+1 merge it

domfarolino:
@JosephPecoraro Personally I like `= "default"` a bit better, but I'm curious what would the label look like for a call to `count()` that would update a counter unique to the callsite (source:line:column)? Something more friendly than `source:line:column : countNumHere`?

Issue #94, opened 2017-03-12
Grouping methods


domfarolino:
The mechanics of the console [grouping methods](https://console.spec.whatwg.org/#grouping) `group()`, `groupEnd()`, and `groupCollapsed()` are pretty straightforward and fairly consistent between implementations (have not tested edge) with the exception of Node (not implemented) and Firefox (shipping collapsible groups in [v54](https://bugzilla.mozilla.org/show_bug.cgi?id=1088360#c35)).

I'm concerned with how Printing is done inside of a group, and how we might describe it. While thinking about how I might implement this in the reference implementation, I came up with something nearly identical to [this](https://github.com/Rich-Harris/node-console-group/blob/master/src/index.js#L7) module which implements the grouping methods (with the exception of `groupCollapsed()`) in Node. It works by maintaining a `prefix` that is modified when grouping functions are called, and is prepended to data before it is printed.

With that being said, I'm thinking we need to include some rhetoric in [Printer](https://console.spec.whatwg.org/#printer) that briefly explains the idea of a prefix that is prepended messages before they get printed. It is probably worth mentioning that the prefix is initially empty. Similarly, for the grouping methods we can include language explaining how the prefix is modified on calls to `group()` and `groupEnd()`. Should we explain that when a group creation method is called, all future Printing is focused inside of a newly created potentially-collapsible representation of an indented group whose members all share the same prefix? The exact characters these methods append to the prefix is *too* specific I think, since it seems to be an implementation-dependent way of indicating indentation:

Chrome:

<img width="288" alt="screen shot 2017-03-12 at 3 38 21 pm" src="https://cloud.githubusercontent.com/assets/9669289/23835227/fafdba00-0739-11e7-9b4c-5bd69293bb16.png">

FF:

<img width="215" alt="screen shot 2017-03-12 at 3 39 40 pm" src="https://cloud.githubusercontent.com/assets/9669289/23835240/3f704d56-073a-11e7-9565-95d88f8c7229.png">

Safari:

<img width="186" alt="screen shot 2017-03-12 at 3 40 27 pm" src="https://cloud.githubusercontent.com/assets/9669289/23835241/44aa20ee-073a-11e7-9f30-5cdaf4c6e1c5.png">

I also propose that if an environment does not support collapsible groups, that the behavior of `groupCollapsed()` is the same as `group()`. This is partially due to historical reasons, as Firefox has done this for pretty much as long as `group()` has been supported, and so that Node can still support calls to this method without throwing.

[...]

PR #95, opened 2017-03-15
Language for grouping methods

domfarolino:
Since I put some group-specific logic inside `Printer` (see "When a group gets printed, output produced by future calls...") should we have a new log level "group" for this to be a little more detectable and rigorous? This would be used inside `group()` and `groupCollapsed()`.

I think the language describing a `group` is both helpful and vague enough to allow browsers to implement this one way, and Node another (possibly even with prefixes since previously outputted data typically is non-interactive) while still being compliant. Thoughts?

Addresses https://github.com/whatwg/console/issues/94

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -141,7 +141,7 @@ The following is an informative summary of the format specifiers processed by th
 
 <h3 id="printer" aoid="Printer" nothrow>Printer(<var>logLevel</var>, <var>args</var>)</h3>
 
-The printer operation is implementation-defined. It accepts a log level indicating severity, and a List of arguments to print (which are either JavaScript objects, of any type, or are implementation-specific representations of printable things such as a stack trace, or output produced by the <code>%o</code> and <code>%O</code> specifiers). How the implementation prints <var>args</var> is up to the implementation, but implementations should separate the objects by a space or something similar, as that has become a developer expectation.
+The printer operation is implementation-defined. It accepts a log level indicating severity, and a List of arguments to print (which are either JavaScript objects, of any type, or are implementation-specific representations of printable things such as a stack trace, a <a>group</a>, or output produced by the <code>%o</code> and <code>%O</code> specifiers). How the implementation prints <var>args</var> is up to the implementation, but implementations should separate the objects by a space or something similar, as that has become a developer expectation.
 
 By the time the printer operation is called, all format specifiers will have been taken into account, and any arguments that are meant to be consumed by format specifiers will not be present in <var>args</var>. The implementation's job is simply to print the List. When a <a>group</a> gets printed, output produced by future calls to Printer should be directed only to the newly printed group.
 
--- a/index.bs
+++ b/index.bs
@@ -143,7 +143,7 @@ The following is an informative summary of the format specifiers processed by th
 
 The printer operation is implementation-defined. It accepts a log level indicating severity, and a List of arguments to print (which are either JavaScript objects, of any type, or are implementation-specific representations of printable things such as a stack trace, or output produced by the <code>%o</code> and <code>%O</code> specifiers). How the implementation prints <var>args</var> is up to the implementation, but implementations should separate the objects by a space or something similar, as that has become a developer expectation.
 
-By the time the printer operation is called, all format specifiers will have been taken into account, and any arguments that are meant to be consumed by format specifiers will not be present in <var>args</var>. The implementation's job is simply to print the List.
+By the time the printer operation is called, all format specifiers will have been taken into account, and any arguments that are meant to be consumed by format specifiers will not be present in <var>args</var>. The implementation's job is simply to print the List. When a <a>group</a> gets printed, ouput produced by future calls to Printer should be directed only to the newly printed group.
 
 If the console is not open when the printer operation is called,
 implementations should buffer messages to show them in the future up to an
@@ -239,7 +239,7 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
 
 <h4 id="clear" oldids="dom-console-clear" method for="console">clear()</h4>
 
-If possible for the environment, clear the console. Otherwise, do nothing.
+End all open groups, and if possible for the environment, clear the console. Otherwise, do nothing.
 
 <h4 id="count" oldids="count-label,dom-console-count" method for="console">count(<var>label</var>)</h4>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -239,7 +239,7 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
 
 <h4 id="clear" oldids="dom-console-clear" method for="console">clear()</h4>
 
-If possible for the environment, clear the console. Otherwise, do nothing.
+End all open groups, and if possible for the environment, clear the console. Otherwise, do nothing.
```
domenic: `<a>groups</a>` I think

index.bs
```diff
@@ -287,12 +287,30 @@ Perform Logger("warn", <var>data</var>).
 
 <h3 id="grouping">Grouping methods</h3>
 
+A <dfn>group</dfn> is an implementation-specific, potentially-interactive view for output produced by calls to Printer, with one further level of indentation than its parent. A group may host output produced by calls to Printer until it is ended with <a>groupEnd()</a>, and only one group will host output produced by calls to Printer at a time.
```
domenic: `<a>groupEnd()</a>` should be `{{console/groupEnd()}}` to get the right formatting.

index.bs
```diff
@@ -287,12 +287,30 @@ Perform Logger("warn", <var>data</var>).
...
+
 <h4 id="group" oldids="group-data,dom-console-group" method for="console">group(...<var>data</var>)</h4>
 
+<emu-alg>
+  1. Let _group_ be a new <a>group</a>.
+  1. Let _groupLabel_ be the result of Formatter(data), and incorporate _groupLabel_ as a label for _group_.
+  1. Optionally if the environment supports interactive groups, _group_ should be expanded by default.
```
domenic: Nit: comma after "Optionally"

index.bs
```diff
@@ -143,7 +143,7 @@ The following is an informative summary of the format specifiers processed by th
 
 The printer operation is implementation-defined. It accepts a log level indicating severity, and a List of arguments to print (which are either JavaScript objects, of any type, or are implementation-specific representations of printable things such as a stack trace, or output produced by the <code>%o</code> and <code>%O</code> specifiers). How the implementation prints <var>args</var> is up to the implementation, but implementations should separate the objects by a space or something similar, as that has become a developer expectation.
```
domenic: I think the list of things that arguments can contain (in parenthesis) needs to be expanded to include the group concept.

index.bs
domfarolino: Nice. I didn't realize `<a>groups</a>` would behave like `<a>group</a>`

domfarolino:
Ah I almost went with a group stack but didn't know if it would be a bit much! I'll put something together referring to a group stack. Should this "global" group stack be first introduced in the text defining what a `group` is, or do you think it would be better somewhere in the top "abstract" area?

domenic:
Hmm interesting question. I guess I might make it something associated with each `console` namespace object?

domfarolino:
> associated with each `console` namespace object?

Sorry, could you elaborate a bit? Are you saying it can be written in IDL under the `console` namespace or am I missing the idea?

In general I guess I'm not really sure how to write something that is properly "global" to the other operations. I naturally lean toward putting it near the abstract operations area because things like `Printer()` and `Logger()` are exposed everywhere without being expressed in IDL.

domenic:
> Sorry, could you elaborate a bit? Are you saying it can be written in IDL under the console namespace or am I missing the idea?

Oh, no, I meant just something like "each `console` namespace object has an associated **group stack**, which is a [stack](https://infra.spec.whatwg.org/#stacks) (initially empty)."

You're right that Printer and Logger are kind of global. Maybe they should take the console object as a parameter. After all if you were implementing this you'd need some way to make sure that Printer() sends things to the right instance of developer tools.

Alternately you can just say something like "the appropriate group stack" and keep things vague. That seems fine too IMO.

domfarolino:
For simplicity I opted to use "appropriate group stack" instead of modify all the abstract ops to take in an instance of the correct `console` namespace object. Perhaps this should be looked into further at some point.

If the above changes are good, I guess the only thing left is recommending that the default group label not be an empty string if `...data` is empty.

index.bs
```diff
@@ -287,12 +287,33 @@ Perform Logger("warn", <var>data</var>).
...
+the last <a>group</a> in a <a>group stack</a> will host output produced by calls to Printer.
+
 <h4 id="group" oldids="group-data,dom-console-group" method for="console">group(...<var>data</var>)</h4>
 
+<emu-alg>
+  1. Let _group_ be a new <a>group</a>.
+  1. Let _groupLabel_ be the result of Formatter(data), and incorporate _groupLabel_ as a label for _group_.
+  1. Optionally, if the environment supports interactive groups, _group_ should be expanded by default.
+  1. Perform Printer("log", «_group_»).
+  1. Push _group_ onto the appropriate <a>group stack</a>
```
domenic: Missing period. Also you can do `<a>Push</a>` (maybe `<a for="stack">Push</a>` will be necessary). You can feel free to not link to such basic infra terms if you think it's silly though; that's up to editor discretion.

index.bs
```diff
@@ -287,12 +287,33 @@ Perform Logger("warn", <var>data</var>).
...
+</emu-alg>
+
 <h4 id="groupcollapsed" oldids="groupcollapsed-data,dom-console-groupcollapsed" method for="console">groupCollapsed(...<var>data</var>)</h4>
 
+<emu-alg>
+  1. Let _group_ be a new <a>group</a>.
+  1. Let _groupLabel_ be the result of Formatter(data), and incorporate _groupLabel_ as a label for _group_.
+  1. Optionally, if the environment supports interactive groups, _group_ should be collapsed by default.
+  1. Perform Printer("log", «_group_»).
+  1. Push _group_ onto the appropriate <a>group stack</a>
```
domenic: Same as above on both counts

index.bs
```diff
@@ -287,12 +287,33 @@ Perform Logger("warn", <var>data</var>).
...
+</emu-alg>
+
 <h4 id="groupend" oldids="dom-console-groupend" method for="console">groupEnd()</h4>
 
+Pop the last <a>group</a> from the <a>group stack</a>.
```
domenic: Can optionally link to pop

index.bs
```diff
@@ -239,7 +239,7 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
 
 <h4 id="clear" oldids="dom-console-clear" method for="console">clear()</h4>
 
-If possible for the environment, clear the console. Otherwise, do nothing.
+End all open <a>groups</a>, and if possible for the environment, clear the console. Otherwise, do nothing.
```
domenic: "End all open groups" should probably be "empty the appropriate group stack" right?

domfarolino: Yep good call

index.bs
domfarolino: I could not get Push/Pop to link to infra without `spec="infra"` (found after skimming bikeshed docs). This seems like overkill though; is this normal especially since `Stack` was picked up so easily?

domenic: Oh, that is an infra bug; the terms are not exported for some reason. Will fix, although it'll be ~24 hours before it gets reindexed afterward, so no need to block this PR on it.

domenic: Posted https://github.com/whatwg/infra/pull/72

domfarolino:
Thanks!

PR #96, opened 2017-03-23
Add dir + dirxml to IDL and spec

domfarolino:
This adds language for `dir` and `dirxml` to the spec.

Whenever we call `Printer` in the spec it seems that we've already made the necessary transformations or conversions ourselves to the objects we're about to pass into `Printer`. To avoid duplication, `group` was defined outside of the methods that use it, therefore anything needing to print a group just says new `group` instead of new `view to host the output produced......`.

In `dirxml` I reused the `%o` format specifier description, and in `dir` I reused portions of the `%O` format specifier description; so I'm wondering if we should define things like:

 - `Generic JavaScript object representation`
and
 - `Optimally useful representation` (or similarly named)

somewhere near Printer so in the future we could say things like

 - "Let *converted* be a **generic JavaScript object** representation of *item*
or
 - "Let *converted* be *item* with **optimally useful formatting** applied.

Then for the two bullet-pointed terms above we could just pull some of the common words out between the definitions that exist already exist in the spec [here](https://console.spec.whatwg.org/#formatter) and [here](https://console.spec.whatwg.org/#formatting-specifiers).

Thoughts?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -195,7 +195,7 @@ namespace console { // but see namespace object requirements below
   void table(any tabularData, optional sequence&lt;DOMString> properties);
   void trace(any... data);
   void warn(any... data);
-  void dir(any data);
+  void dir(any item);
   void dirxml(any... data);
 
   // Grouping
@@ -287,11 +287,11 @@ Try to construct a table with the columns of the properties of <var>tabularData<
 
 Perform Logger("warn", <var>data</var>).
 
-<h4 id="dir" oldids="dir-data,dom-console-dir" method for="console">dir(<var>data</var>)</h4>
+<h4 id="dir" method for="console">dir(<var>item</var>)</h4>
 
-Let <var>object</var> be an potentially-interactive representation of <var>data</var> treated as a JavaScript object. Perform Printer("log", <var>«object»</var>).
+Let <var>object</var> be an potentially-interactive representation of <var>item</var> treated as a JavaScript object. Perform Printer("log", <var>«object»</var>).
 
-<h4 id="dirxml" oldids="dirxml-data,dom-console-dirxml" method for="console">dirxml(...<var>data</var>)</h4>
+<h4 id="dirxml" method for="console">dirxml(...<var>data</var>)</h4>
 
 <emu-alg>
   1. Let _finalList_ be a new <a>list</a>, initially empty.
--- a/index.bs
+++ b/index.bs
@@ -195,6 +195,8 @@ namespace console { // but see namespace object requirements below
   void table(any tabularData, optional sequence&lt;DOMString> properties);
   void trace(any... data);
   void warn(any... data);
+  void dir(any data);
+  void dirxml(any... data);
 
   // Grouping
   void group(any... data);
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -285,6 +287,18 @@ Try to construct a table with the columns of the properties of <var>tabularData<
 
 Perform Logger("warn", <var>data</var>).
 
+<h4 id="dir" oldids="dir-data,dom-console-dir" method for="console">dir(<var>data</var>)</h4>
+
+Let <var>object</var> be an potentially-interactive representation of <var>data</var> treated as a JavaScript object. Perform Printer("log", <var>«object»</var>).
```
domenic: This should be two separate steps.

index.bs
```diff
@@ -285,6 +287,18 @@ Try to construct a table with the columns of the properties of <var>tabularData<
 
 Perform Logger("warn", <var>data</var>).
 
+<h4 id="dir" oldids="dir-data,dom-console-dir" method for="console">dir(<var>data</var>)</h4>
```
domenic: No need for oldids in new sections :)

index.bs
```diff
@@ -195,6 +195,8 @@ namespace console { // but see namespace object requirements below
   void table(any tabularData, optional sequence&lt;DOMString> properties);
   void trace(any... data);
   void warn(any... data);
+  void dir(any data);
```
domenic: Should we perhaps use a different variable name for single arguments, leaving "data" for `any...` cases? I'm not sure I have any good suggestions though. Maybe `datum`? Meh...

index.bs
```diff
@@ -285,6 +287,18 @@ Try to construct a table with the columns of the properties of <var>tabularData<
...
 
+<h4 id="dir" oldids="dir-data,dom-console-dir" method for="console">dir(<var>data</var>)</h4>
+
+Let <var>object</var> be an potentially-interactive representation of <var>data</var> treated as a JavaScript object. Perform Printer("log", <var>«object»</var>).
+
+<h4 id="dirxml" oldids="dirxml-data,dom-console-dirxml" method for="console">dirxml(...<var>data</var>)</h4>
+
+<emu-alg>
+  1. Let _finalList_ be a new <a>list</a>, initially empty.
+  1. For each _item_ of _data_, let _converted_ be a DOM tree representation of _item_ if possible, otherwise let _converted_ be an implementation-specific representation of _item_ judged to be maximally useful and informative. Append _converted_ to _finalList_.
```
domenic: I would probably have this as two nested steps under the for each. Up to you though.

index.bs
```diff
@@ -285,6 +287,18 @@ Try to construct a table with the columns of the properties of <var>tabularData<
 
 Perform Logger("warn", <var>data</var>).
 
+<h4 id="dir" oldids="dir-data,dom-console-dir" method for="console">dir(<var>data</var>)</h4>
+
+Let <var>object</var> be an potentially-interactive representation of <var>data</var> treated as a JavaScript object. Perform Printer("log", <var>«object»</var>).
+
+<h4 id="dirxml" oldids="dirxml-data,dom-console-dirxml" method for="console">dirxml(...<var>data</var>)</h4>
+
+<emu-alg>
```
domenic: This algorithm doesn't seem to process formatting specifiers, but Chrome does and Firefox does not, hmm. Maybe the best path is to leave as-is in this PR but open a separate issue to discuss?

index.bs
domfarolino: What exactly are they for (or residual from) anyways? Can they be removed from the `group` section too without harming things?

domfarolino: I'm assuming it is literally a list of old-ids that other specs may refer to, so we tell bikeshed "Hey this thing may also go by these names too {name1, name2, ...}"?

index.bs
domfarolino: Went with `item` here, since in `dirxml` when I iterate over `data` I call individual elements `item`s

domfarolino:
I'm going to go ahead and add the `<dfn>`s to Printer in this PR too in a bit

index.bs
domfarolino: That is reasonable. Must've slipped my mind. I think in the end I lean toward processing format specifiers just as `log` does, because `dirxml` behaves so much like `log`

domfarolino:
Ahh, actually no I'll make that separate because that'll end up modifying both "Formatter" and "Summary of formatting specifiers" sections. I'll do a separate PR to address #97

index.bs
domenic: Yes, exactly; they were IDs that used to exist for those methods in the document before some restructuring/renaming. It looks like they were introduced in https://github.com/whatwg/console/commit/64a3ff6df3456daa5cc50efe3cd8e7141d4ea402; previously each heading had two separate anchors (one for the h4 and one for the dfn), both of which had suboptimal IDs. That commit gave everything optimal IDs but then introduced oldids so as to not break any inbound links.

index.bs
```diff
@@ -285,6 +287,23 @@ Try to construct a table with the columns of the properties of <var>tabularData<
 
 Perform Logger("warn", <var>data</var>).
 
+<h4 id="dir" method for="console">dir(<var>item</var>)</h4>
+
+<emu-alg>
+  1. Let <var>object</var> be an potentially-interactive representation of <var>item</var> treated as a JavaScript object.
+  1. Perform Printer("log", <var>«object»</var>).
```
domenic: The angled bracket thingies don't belong inside the `<var>` tags

PR #100, opened 2017-03-25
Add Formatter functionality to dirxml

domfarolino:


domfarolino:
@domenic this is just a trivial change to close #100 , look ok?

index.bs
```diff
@@ -301,7 +301,7 @@ Perform Logger("warn", <var>data</var>).
   1. For each _item_ of _data_:
     1. Let _converted_ be a DOM tree representation of _item_ if possible, otherwise let _converted_ be an implementation-specific representation of _item_ judged to be maximally useful and informative.
     1. Append _converted_ to _finalList_.
-  1. Perform Printer("log", _finalList_).
+  1. Perform Printer("log", Formatter(_finalList_)).
```
domenic: Hmm, is this better or worse than Logger("Log", _finalList_)? For example I think Formatter expects at least two args. (It should probably get an assert about that...)

domfarolino: Good point, calling Formatter inside Printer is pretty much just calling Logger, will change. Yeah Formatter could probably be a little more rigorous

PR #101, opened 2017-03-26
Factor out object formatting definitions

### LINKED ISSUE #97: Pull %O and %o descriptions out into dfns inside Printer for reuse
As briefly discussed in https://github.com/whatwg/console/pull/96

domfarolino:
First go at factoring out definitions we've duplicated in the spec when referring to the Printing of objects. Will close #97.

Originally I was going to put the dfns in the main Printer section but decided that just mashing them in between paras might not be the best way to keep that section clean.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -141,7 +141,7 @@ The following is an informative summary of the format specifiers processed by th
 
 <h3 id="printer" aoid="Printer" nothrow>Printer(<var>logLevel</var>, <var>args</var>)</h3>
 
-The printer operation is implementation-defined. It accepts a log level indicating severity, and a List of arguments to print (which are either JavaScript objects, of any type, or are implementation-specific representations of printable things such as a stack trace, a <a>group</a>, or output produced by the <code>%o</code> and <code>%O</code> specifiers). How the implementation prints <var>args</var> is up to the implementation, but implementations should separate the objects by a space or something similar, as that has become a developer expectation.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -149,6 +149,12 @@ If the console is not open when the printer operation is called,
 implementations should buffer messages to show them in the future up to an
 implementation-chosen limit (typically on the order of at least 100).
 
+<h4 id="object-formats">Common Object Formats</h4>
+
+Typically objects will be printed in a format that is suitable for their context. This section describes common ways in which objects are formatted to be most useful in their context.
+
+An object with <dfn>generic JavaScript object formatting</dfn> is a potentially expandable representation of a generic JavaScript object. An object with <dfn>optimally useful formatting</dfn> is an implementation-specific, potentially-interactive object judged to be maximally useful and informative.
```
domenic: I'd go with "representation of an object" instead of "object" for the last clause here.

index.bs
```diff
@@ -149,6 +149,12 @@ If the console is not open when the printer operation is called,
 implementations should buffer messages to show them in the future up to an
 implementation-chosen limit (typically on the order of at least 100).
 
+<h4 id="object-formats">Common Object Formats</h4>
+
+Typically objects will be printed in a format that is suitable for their context. This section describes common ways in which objects are formatted to be most useful in their context.
```
domenic: Oh, and it may be useful in this section to kind of expand on the idea that these are implementation-specific representations which will eventually end up Printer, which is where they have actual importance.

PR #102, opened 2017-03-27
Meta: add .editorconfig and .gitattributes

domenic:
Part of https://github.com/whatwg/meta/issues/7.

I *believe* this will also instruct GitHub to wrap lines in .bs files by treating them as .html files. That should help reviewing a lot.

Alternately the spec could start wrapping at 100 characters like most WHATWG specs; not sure if that's worthwhile/what the editors want though.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_size = 2
+indent_style = space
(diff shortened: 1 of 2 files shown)


domfarolino:
A wrap is definitely what I want. I was actually going to ask on IRC if there was a norm for this as I was curious why this spec had massive run-on lines.

domenic:
Great. Let's leave wrapping as separate. If we do let's remember to add the line to editorconfig with the max line length,

Issue #90, opened 2017-03-03
The spec for console.count() implies it should log every time


domenic:
https://github.com/whatwg/console/issues/60 discusses an issue with the current spec for console.count(), which says

> Perform Logger("log", concat).

But given code like

```js
console.count("a");
console.count("a");
```

implementations will generally not log

```
a: 1
a: 2
```

but instead will update the first line to say "a: 2".

----

Now, since Printer (called by Logger) is implementation defined, the spec actually allows this just fine. But it's pretty unclear. It's also a bit strange since we do `Logger("log", concat)`, so the implementation would have to use some out-of-band magic to remember that this `"log"` is not the same as `console.log()`s `"log"`.

Personally I think the best fix would be:

- Introduce a new log level, "count", and use it in the definition of console.count()
- Add a sentence in the description of Printer that talks about what implementations often do for a log level of "count".

domfarolino:
A tad confused here...from what I can tell copying

```js
console.count("a");
console.count("a");
```

into the DevTools gives

```js
a: 1
a: 2
```

in Chrome + Canary, Safari + Tech Preview, and Firefox 52. I vaguely remember it was Edge that was prone to updating a previously "logged" counter when calling `count(existingLabel)` from a previous discussion. With that being said, is this just an edge thing or am I missing something?

domenic:
Huh, for some reason I could swear I observed the counter-updating behavior in some browser. But re-reading #60 it looks like @JosephPecoraro was mainly suggesting that it be allowed, not that browsers already do it. I guess that decreases the urgency here...

JosephPecoraro:
From https://github.com/whatwg/console/issues/60#issuecomment-217998248:

> In Edge you should see the UI update just one entry in the console when console.count is called and not add an entry for every increment.

xirzec:
I was actually going to change Edge to behave like Chrome/Safari/FireFox and log each time to be consistent. Is this not desirable?

domfarolino:
Personally I like the consistency. I think the reason the issue was raised was because we noticed the odd behavior with Edge. I am in favor of this changing for consistency though, thoughts from anyone else?

JosephPecoraro:
It still feels to me like things are being over-specified without actually thinking through what would actually be useful for developers.

I view `console.count` as being about incrementing a counter and having developer tools provide a way to view the value of the counter.

Things that would need to be defined are:

  * How the counters are created / named / shared
  * How the counters can be incremented, decremented, or cleared

[...]

PR #105, opened 2017-03-27
Make the spec for time()/timeEnd() a bit more accurate and rigorous

domenic:
Notable changes include the addition of a colon to the output, being explicit about how the implementation decides the format of the duration string, and making the timer table concept explicit.

This also fixes up a few minor linking errors, e.g. two elements having the same ID "group".

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -17,7 +17,11 @@ Logo: https://resources.whatwg.org/logo-console.svg
 !Commits: <a href="https://twitter.com/consolelog">@consolelog</a>
 
 Opaque Elements: emu-alg
-Link Defaults: html (dfn) structured clone
+</pre>
+
+<pre class="link-defaults">
+spec:infra; type:dfn; text:list
+spec:infra; type:dfn; text:string
 </pre>
 
 <style>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -306,7 +310,7 @@ Perform Logger("warn", <var>data</var>).
 
 <h3 id="grouping">Grouping methods</h3>
 
-A <dfn>group</dfn> is an implementation-specific, potentially-interactive view for output produced by calls to Printer, with one further level of indentation than its parent. Each {{console}} namespace object has an associated <dfn>group stack</dfn>, which is a <a>stack</a>, initially empty. Only
+A <dfn id="concept-group">group</dfn> is an implementation-specific, potentially-interactive view for output produced by calls to Printer, with one further level of indentation than its parent. Each {{console}} namespace object has an associated <dfn>group stack</dfn>, which is a <a>stack</a>, initially empty. Only
```
domfarolino: Does this prevent name collision with the `group` method?

domenic:
@domfarolino rebased this

domfarolino:
@domenic This looks good thanks a lot! My only question is regarding the usage of "info" from my above comment, but as stated we can probably just address https://github.com/whatwg/console/issues/90#issuecomment-289260322 (fancy custom UI per console func) separately; wanted to run it by you though.

domenic:
Sorry, I can't see your above comment. Yeah, addressing that separately makes sense to me; this preserves the already-existing usage of "info" for now.

domfarolino:
Odd, gotcha. I was just curious about the already-existing usage of "info". A picture I included in the elusive review comment was basically:
screen shot 2017-04-03 at 10 00 35 pm

showing that implementations seem to use some custom "info" however since we'll be taking care of it elsewhere, makes no difference

refack:
I wanted to ask what happens when `label` is not a string. This PR does not call for a `ToString` conversion, so:
1. Should it throw for non strings?
2. Should it do a `ToString` conversion?
3. Should it coarse `label` to string with `String('label')`?

domenic:
This specification defines

```webidl
  // Timing
```

so your answers are [found in the Web IDL specification](https://heycam.github.io/webidl/#es-DOMString).

refack:
@domenic thank you!

Issue #79, opened 2016-12-19
Node.js console.dir


Fishrock123:
Node.js has implemented a [`console.dir(obj[, options])`](https://nodejs.org/dist/latest-v7.x/docs/api/console.html#console_console_dir_obj_options) since Node v0.1.101.

From what I can tell, it appears to be a more direct sort of proxy to [`util.format()`](https://nodejs.org/dist/latest-v7.x/docs/api/util.html#util_util_inspect_object_options) allowing the passing of options rather than additional objects.

domenic:
dir is tracked in https://github.com/whatwg/console/issues/27 FWIW. Someone should really fix that... slapping "good first bug" on it doesn't seem to be working :)

domfarolino:
This can be closed since `dir` has been added but I noticed browser implementations don't seem to support the optional `options` object that Node does. It also doesn't seem really applicable to browser implementations but I'm wondering if the API difference matters at all, or if we should at least make a note about it in the spec?

domenic:
Hmm. This is a case where I think it doesn't hurt anyone to include special optional Node.js-only provisions. E.g. an optional argument for the options which anyone can use to customize the behavior of Printer, but in practice only Node.js uses this flexibility.

Based on [the source code](https://github.com/nodejs/node/blob/ed12ea371c5352b4b13f9f1c4e0f577fbd30bb2a/lib/console.js#L126) in Node plus [the definition of Object.assign](https://tc39.github.io/ecma262/#sec-object.assign) it looks like IDL of `optional object? options` would be equivalent. I believe such IDL cannot throw any exception, and I know it will not impact `console.dir.length`, so browsers that totally ignore the argument will be compliant.

Then we could modify step 2 to say something vague like "optionally using the properties of the _options_ object to customize display". Or we could add a third arg to Printer, pass it through, and add a sentence about optionally using the third arg (if it is not undefined or null) to customize the display; that might be cleaner.

The only possible downside of this is that it constrains browsers from ever using the second argument in some different way. But that doesn't seem like a realistic fear.

domfarolino:
Yes I favor the third arg over the method doing stuff that is in Printer's domain. The optional + nullable options object sounds good thanks for digging in. Going to reopen this until I take care of it.

PR #107, opened 2017-03-30
Support Nodejs dir

domfarolino:
Will close: https://github.com/whatwg/console/issues/79.

I am not sure if the `h3` Printer header will need any special symbols on the `options` object to indicate that it is optional and thus make all other calls to Printer with two arguments valid.

This also addresses a small typo I missed when adding text about the group stack in Printer.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -153,22 +153,31 @@ The following is an informative summary of the format specifiers processed by th
   </tr>
 </table>
 
-<h3 id="printer" aoid="Printer" nothrow>Printer(<var>logLevel</var>, <var>args</var>)</h3>
+<h3 id="printer" aoid="Printer" nothrow>Printer(<var>logLevel</var>, <var>args</var>, <var>options</var>)</h3>
 
-The printer operation is implementation-defined. It accepts a log level indicating severity, and a
-List of arguments to print (which are either JavaScript objects, of any type, or are
-implementation-specific representations of printable things such as a stack trace, a <a>group</a>,
-or objects with either <a>generic JavaScript object formatting</a> or
-<a>optimally useful formatting</a> applied). How the implementation prints <var>args</var> is up to
-the implementation, but implementations should separate the objects by a space or something
-similar, as that has become a developer expectation.
+The printer operation is implementation-defined. It accepts a log level indicating severity, a List
+of arguments to print, and an optional object of implementation-specific formatting options.
+Elements appearing in <var>args</var> will be one of the following:
+
+<emu-alg>
+  1. JavaScript objects of any type.
+  1. Implementation-specific representations of printable things such as a stack trace or
+  <a>group</a>.
+  1. Objects with either <a>generic JavaScript object formatting</a> or
+  <a>optimally useful formatting</a> applied.
+</emu-alg>
+
+Optionally, if the <var>options</var> object is not undefined, implementations can use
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -153,22 +153,31 @@ The following is an informative summary of the format specifiers processed by th
...
-implementation-specific representations of printable things such as a stack trace, a <a>group</a>,
-or objects with either <a>generic JavaScript object formatting</a> or
-<a>optimally useful formatting</a> applied). How the implementation prints <var>args</var> is up to
-the implementation, but implementations should separate the objects by a space or something
-similar, as that has become a developer expectation.
+The printer operation is implementation-defined. It accepts a log level indicating severity, a List
+of arguments to print, and an optional object of implementation-specific formatting options.
+Elements appearing in <var>args</var> will be one of the following:
+
+<emu-alg>
```
domenic: I think this will be better as a `<ul>`, not an algorithm. I believe you should be able to use Markdown-like syntax for that.

index.bs
```diff
@@ -153,22 +153,31 @@ The following is an informative summary of the format specifiers processed by th
   </tr>
 </table>
 
-<h3 id="printer" aoid="Printer" nothrow>Printer(<var>logLevel</var>, <var>args</var>)</h3>
+<h3 id="printer" aoid="Printer" nothrow>Printer(<var>logLevel</var>, <var>args</var>, <var>options</var>)</h3>
```
domenic: Traditionally you do `[, options]` to denote optionality.

index.bs
```diff
@@ -153,22 +153,31 @@ The following is an informative summary of the format specifiers processed by th
...
+
+<emu-alg>
+  1. JavaScript objects of any type.
+  1. Implementation-specific representations of printable things such as a stack trace or
+  <a>group</a>.
+  1. Objects with either <a>generic JavaScript object formatting</a> or
+  <a>optimally useful formatting</a> applied.
+</emu-alg>
+
+Optionally, if the <var>options</var> object is not undefined, implementations can use
```
domenic: I think this needs to say something like "If the options object is passed, and is not undefined or null, ..."

index.bs
```diff
@@ -153,22 +153,31 @@ The following is an informative summary of the format specifiers processed by th
...
+<var>options</var> to apply implementation-specific formatting to the elements in <var>args</var>.
+How the implementation prints <var>args</var> is up to the implementation, but implementations
```
domenic: I think this is nicer as a second paragraph. Once-sentence paragraphs are reasonably common in specs.

index.bs
domfarolino: Good call, totally overlooked basic HTML

index.bs
domfarolino: Odd, pretty sure I addressed this in my commits though it didn't collapse here

index.bs
```diff
@@ -153,22 +153,30 @@ The following is an informative summary of the format specifiers processed by th
   </tr>
 </table>
 
-<h3 id="printer" aoid="Printer" nothrow>Printer(<var>logLevel</var>, <var>args</var>)</h3>
+<h3 id="printer" aoid="Printer" nothrow>Printer(<var>logLevel</var>, <var>args</var>, [<var>, options</var>])</h3>
```
domenic: Double commas

robertkowalski:
wow great work

<3

PR #109, opened 2017-05-09
Elaborate on Printer UX and log level severity

domfarolino:
My first take on https://github.com/whatwg/console/issues/90

The PR
 - Fixes a typo
 - Fixes some HTML table semantic inconsistencies appearing in the spec
 - Customizes all calls to Printer and Logger by not reusing the same `logLevels`
 - Adds a section/table discussing/grouping console methods by log level severity
 - Adds a section discussing some possible and common UX innovations Printer may perform
 - should close https://github.com/whatwg/console/issues/90
 - that might be it?

@domenic Note the TODO section appearing in the "Indicating logLevel Severity" section. If you want to push up a picture of an example that the TODO is talking about (edge), that might be helpful to include.


index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
 implementations should buffer messages to show them in the future up to an
 implementation-chosen limit (typically on the order of at least 100).
 
+<h4 id="loglevel-severity">Indicating <var>logLevel</var> Severity</h4>
+
+This section groups together {{console}} methods whose <var>logLevel</var>s produce output
+indicating similar levels of severity.
+
+<div class="example" id="loglevel-example">
```
terinjokes: This should probably be a note, not an example.

domfarolino: Hmm, I'm rethinking the usefulness of this note/example anyways. Do you think it is necessary? I think the table might be clear enough to warrant getting rid of this. Thoughts...

index.bs
```diff
@@ -126,39 +126,38 @@ The following is an informative summary of the format specifiers processed by th
...
   <tr>
     <td><code>%c</code></td>
-    <td>Applies provided CSS</<td>
+    <td>Applies provided CSS</td>
     <td>n/a</td>
   </tr>
 </table>
 
-<h3 id="printer" aoid="Printer" nothrow>Printer(<var>logLevel</var>, <var>args</var> [<var>, options</var>])</h3>
+<h3 id="printer" aoid="Printer" nothrow>Printer(<var>logLevel</var>, <var>args</var> [<var ignore=''>, options</var>])</h3>
```
domenic: The `<var>` should be around `options`, not including the comma

index.bs
domenic: I agree this isn't necessary.

index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
...
+<div class="example" id="loglevel-example">
+  Since <code>log</code> and <code>debug</code> produce similarly formatted outputs, they are
+  grouped together by their log level severity in the following table.
+</div>
+
+<table>
+  <thead>
+    <tr>
+      <th>{{console}} method</th>
+      <th>Log Level Severity</th>
```
domenic: Just "Log Level"?

index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
...
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tr>
+    <td><code>log, trace, dir, dirxml, group, groupCollapsed, debug</code></td>
+    <td>log</td>
+    <td>
+      Console message with implementation-specific formatting designed to indicate a generic log
```
domenic: The common prefix "Console message with implementation-specific formatting designed to" is a bit annoying and can probably be removed

index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
...
+<table>
+  <thead>
+    <tr>
+      <th>{{console}} method</th>
+      <th>Log Level Severity</th>
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tr>
+    <td><code>log, trace, dir, dirxml, group, groupCollapsed, debug</code></td>
```
domenic: These should be done as

```
{{console/log()}}, {{console/trace()}}, ...
```

to get auto-linking and better formatting.

index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
...
+</table>
+
+While the above table groups together {{console}} methods by <var>logLevel</var> severity, the
+Printer is still implementation-defined, and is therefore allowed to provide formatting variations
+on a per-method basis while remaining compliant. Some common examples of this UX improvisation are
+as follows:
+
+<ul>
+  <li>
+    <code>timeEnd</code> custom formatting.
```
domenic: Again method references should be done via `{{console/timeEnd()}}`

index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
...
+    <div class="example" id="timeEnd-formatting">
+      Formatting applied to output produced by calls to <code>timeEnd</code> may be unique to this
+      method.
+
+      <img src="images/timeEnd-formatting.png" />
```
domenic: Images need `alt=""` descriptions, and don't need `/>`, just `>` suffices.

index.bs
domenic: Each bullet should probably be a separate `<div class="example">`, instead of using a bulletted list.

index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
 implementations should buffer messages to show them in the future up to an
 implementation-chosen limit (typically on the order of at least 100).
 
+<h4 id="loglevel-severity">Indicating <var>logLevel</var> Severity</h4>
```
domenic: Headings should not be title case, so "severity" should not be capitalized

index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
...
+      <img src="images/timeEnd-formatting.png" />
+    </div>
+  </li>
+  <li>
+    <code>count</code> does not always print new output.
+    <p class="XXX">TODO: A picture of this in the Edge console would be nice.</p>
+  </li>
+</ul>
+
+<h4 id="printer-ux-innovation">Printer UX Innovation</h4>
```
domenic: Again no titlecase.

"UX" is an uncommon enough acronym in specs that it should probably be spelled out. It also might be clearer to just use the more common "UI" or "UI widgets" for the bullet points.

index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
...
+
+Since Printer is implementation-defined, it is common to see UX innovations in its implementations.
+The following is a non-exhaustive list of UX enhancements that may appear in existing
+implementations:
+
+<ul>
+  <li>
+    De-duplication of identical output to prevent spam.
```
domenic: `<p>` tags around the list item contents that are not `<div class="example">` will probably make this look a little nicer.

index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
...
+    <div class="example" id="dedupe-example">
+      An implementation may choose to batch multiple, identical messages together to avoid spamming
+      the console. It may also be helpful to provide the number of messages that have been batched
+      together in a single entry.
```
domenic: This paragraph is a bit redundant with the previous one. Maybe "In this example, the implementation not only batches together multiple identical messages, but also provides the number of messages that have been batched together".

index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
...
+  </li>
+  <li>
+    Extra UX off to the side allowing the user to filter messages by log level severity.
+    <div class="example" id="severity-filter-example">
+      <img src="images/severity-filter.png" />
+    </div>
+  </li>
+  <li>
+    Extra UX off to the side indicating the current state of the <a>timer table</a>,
+    <a>group stack</a>, or some other internally maintained data.
```
domenic: s/some other/other

index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
...
+  </li>
+  <li>Flashing portions of the console to alert the user of something important.</li>
+</ul>
+
 <h4 id="object-formats">Common Object Formats</h4>
```
domenic: Might as well un-titlecase this while there.

index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
 implementations should buffer messages to show them in the future up to an
 implementation-chosen limit (typically on the order of at least 100).
 
+<h4 id="loglevel-severity">Indicating <var>logLevel</var> Severity</h4>
+
+This section groups together {{console}} methods whose <var>logLevel</var>s produce output
+indicating similar levels of severity.
```
domenic: This paragraph confusing because _logLevel_ is now unique per-method, whereas the "Log Level Severity" column is a grouping. Similarly the paragraph afterward doesn't make much sense to me.

Here would be my take:

> Each {{console}} method uses a unique value for the _logLevel_ parameter when calling Printer, allowing implementations to customize each printed message depending on the method from which it originated. However, it is common practice to group together certain methods and treat their output similarly, in four broad categories. This table summarizes these common groupings:

(the table goes here. Maybe the "Log Level Severity" column gets renamed "Grouping" and goes first.)

> These groupings are meant to document common practices, and do not constrain implementations from providing special behavior for each method, as in the following examples:

index.bs
```diff
@@ -187,6 +186,108 @@ If the console is not open when the printer operation is called,
...
+Printer is still implementation-defined, and is therefore allowed to provide formatting variations
+on a per-method basis while remaining compliant. Some common examples of this UX improvisation are
+as follows:
+
+<ul>
+  <li>
+    <code>timeEnd</code> custom formatting.
+    <div class="example" id="timeEnd-formatting">
+      Formatting applied to output produced by calls to <code>timeEnd</code> may be unique to this
+      method.
```
domenic: Since it's not clear exactly what's unique about timeEnd, it may be better to expand this example to include both console.timeEnd() and console.info(), and explain in prose something like "here you can see an implementation has chosen to make timeEnd()'s output blue, while leaving the output for info() a neutral color."

index.bs
domfarolino: Ah, couldn't remember the formatting for that..perfect

index.bs
domfarolino: Gotcha...honestly I just copied it from another `img` tag in the spec, so I'll touch it up too

domfarolino: @domenic Do you mean they must specifically have *empty* alternative text, or that they just *need* some alternate text?

domenic: Some alternate text, although in some cases empty might be most appropriate. https://html.spec.whatwg.org/multipage/embedded-content.html#alt

index.bs
domfarolino: While expanding this example I found that Chrome 58 gives this:

latest

...basically the `info` formatting and its variants have disappeared. Do you think we should reconsider the spec and possible get rid of the `info` log level, or just include a picture of another implementation and maybe file a bug (though I'm not sure if that'll is technically necessary) on Chrome?

I tried to checkout the source to understand what was going on or when this change was made. Both
 - https://cs.chromium.org/chromium/src/v8/src/inspector/v8-console-message.cc?type=cs&l=391
 - https://cs.chromium.org/chromium/src/v8/include/v8.h?type=cs&l=6804

indicate that an internal log level of `info` still exists, among others...but I'm really not sure what is going on past that.

index.bs
domfarolino: Sorry, are you saying change "some other" => "other"?

domenic: Yeah

index.bs
domfarolino: I like this, thanks!

index.bs
terinjokes: @domfarolino `console.info` should definitely still exist in this specification, but again, as this section notes, implementations are allowed to format it anyway they want.

terinjokes:
Once you're done with changes, do you mind squashing them locally and pushing a single commit up? I really don't like how GitHub does it with the merge button.

index.bs
domfarolino: Oh yes `console.info` should definitely be in the spec, but I'm wondering if we should remove  the broad `info` section from the grouping table or not...idk just a thought. I guess leaving it wouldn't hurt at all.

terinjokes: @domfarolino I don't see how it hurts. It's a way at least one implementor displayed it, and I don't think we plan on updating this section each time an implementor decides to experiment with UX.

index.bs
```diff
@@ -187,7 +186,101 @@ If the console is not open when the printer operation is called,
...
+
+  <img alt="A demonstration of timeEnd and info formatting differences" src="images/timeEnd-formatting.png">
+</div>
+
+<div class="example" id="count-output">
+  Calls to {{console/count()}} may not always print new output.
+  <p class="XXX">TODO: A picture of this in the Edge console would be nice.</p>
+</div>
+
+<h4 id="printer-ux-innovation">Printer User Experience innovation</h4>
```
domenic: Lowercase "user experience"

index.bs
```diff
@@ -187,7 +186,101 @@ If the console is not open when the printer operation is called,
...
+
+Since Printer is implementation-defined, it is common to see UX innovations in its implementations.
+The following is a non-exhaustive list of UX enhancements that may appear in existing
+implementations:
```
domenic: "may appear in existing implementations" is not great since "may" is supposed to be for normative requirements. Maybe just "non-exhaustive list of potential UX enhancements"?

index.bs
```diff
@@ -187,7 +186,101 @@ If the console is not open when the printer operation is called,
...
+Each {{console}} method uses a unique value for the logLevel parameter when calling Printer,
+allowing implementations to customize each printed message depending on the method from which
+it originated. However, it is common practice to group together certain methods and treat their
+output similarly, in four broad categories. This table summarizes these common groupings:
+
+<table>
+  <thead>
+    <tr>
+      <th>Grouping</th>
+      <th>{{console}} method</th>
```
domenic: methods, not method

index.bs
```diff
@@ -187,7 +186,101 @@ If the console is not open when the printer operation is called,
...
+<div class="example" id="timeEnd-formatting">
+  Formatting applied to output produced by calls to {{console/timeEnd()}} may be unique to this
+  method.
+
+  <img alt="A demonstration of timeEnd and info formatting differences" src="images/timeEnd-formatting.png">
+</div>
+
+<div class="example" id="count-output">
+  Calls to {{console/count()}} may not always print new output.
+  <p class="XXX">TODO: A picture of this in the Edge console would be nice.</p>
```
domenic: Here you go! 
repeated-count-updates

With this, maybe revise the previous sentence to include "but instead could update previously-output counts" or similar.

index.bs
```diff
@@ -187,7 +186,101 @@ If the console is not open when the printer operation is called,
 implementations should buffer messages to show them in the future up to an
 implementation-chosen limit (typically on the order of at least 100).
 
-<h4 id="object-formats">Common Object Formats</h4>
+<h4 id="loglevel-severity">Indicating <var>logLevel</var> severity</h4>
+
+Each {{console}} method uses a unique value for the logLevel parameter when calling Printer,
```
domenic: `<var>` around logLevel

index.bs
domfarolino: Ok I think that makes sense. With that said, I revised the `count` para to say:

> Calls to {{console/count()}} **might** not always print new output, but instead could update previously-output counts.

(not actually bold in spec). I think that *might* be a little better too, since that word isn't mentioned in https://infra.spec.whatwg.org/#conformance

domfarolino:
@domenic Could you just check out the changes I made to the `timeEnd` formatting explanation? It looks like I forgot to address them so I just pushed a change.

@terinjokes I'm assuming you mean `git rebase -i HEAD~N`? I've never done it locally is why I ask,
 just want to make sure we're on the same page. ([source](https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit))

domenic:
LGTM

domfarolino:
Sorry, I couldn't get the squash working locally for some reason.

PR #112, opened 2017-05-30
Clarify console.time() steps

### LINKED ISSUE #103: console.time() multiple times with the same label
The current spec seems to imply this will reset the timer, but at least in some of my tests, it does not seem to. We should get to the bottom of this, and write some (manual) web platform tests.

domfarolino:
Will close #103

@domenic do you think the way I'm handling the possible warning reporting is appropriate or is it maybe too vague? I thought explaining reasonable error messages and explicitly calling Printer("warn", ...) would be a little out of the scope of this function.

Edit: just for some record-keeping: https://bugs.chromium.org/p/chromium/issues/detail?id=727514

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -514,7 +514,7 @@ Each {{console}} namespace object has an associated <dfn>timer table</dfn>, whic
      reporting a warning to the console indicating that a timer with label _label_ has already been
      started.
   1. Otherwise, <a for="map">set</a> the value of the entry with key _label_ in the associated
-     <a>timer table</a> to the current time in an implementation-defined format.
+     <a>timer table</a> to the current time.
 </emu-alg>
 
 <h4 id="timeend" oldids="timeend-label,dom-console-timeend" method for="console">timeEnd(<var>label</var>)</h4>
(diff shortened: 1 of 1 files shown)


terinjokes:
I'm not @domenic, but this looks good to me.

domfarolino:
@terinjokes Great thanks!

index.bs
```diff
@@ -509,8 +509,13 @@ Each {{console}} namespace object has an associated <dfn>timer table</dfn>, whic
...
 <h4 id="time" oldids="time-label,dom-console-time" method for="console">time(<var>label</var>)</h4>
 
-<a for="map">Set</a> the value of the entry with key <var>label</var> in the associated
-<a>timer table</a> to the current time.
+<emu-alg>
+  1. If the associated <a>timer table</a> contains an entry with key _label_, abort, optionally
+     reporting a warning to the console indicating that a timer with label _label_ has already been
+     started.
+  1. Otherwise, <a for="map">set</a> the value of the entry with key _label_ in the associated
+     <a>timer table</a> to the current time in an implementation-defined format.
```
domenic: I'm not sure we need "in an implementation-defined format" since this is a time, not a string.

Issue #114, opened 2017-06-06
Specify infra map in `console.count()`


domfarolino:
This mainly stems from https://github.com/whatwg/console/issues/88#issuecomment-294854936 but yeah, we should probably formalize how we set and retrieve count labels, and [infra's map](https://infra.spec.whatwg.org/#maps) seems like the way to do it.

If someone wanted to take a stab at this, I'd suggest getting a little familiar with that data structure a bit. In this spec specifically, good examples of its usage can be found in the section specifying the [timer methods](https://console.spec.whatwg.org/#timing).

Hmm, maybe @musgravejw wants to have a go?

domfarolino:
Closed by #115

PR #115, opened 2017-06-12
[#114] Specify map for console.count labels

musgravejw:
Addresses https://github.com/whatwg/console/issues/114

Attempt to define `console.count()` label behavior by using an ordered map entry for each label:  https://infra.spec.whatwg.org/#maps

Specifies that a map object should be created, and that calls to count with that label should increment the value associated with the label.  If a label is undefined, a label passed to count will set the value of key _label_ to be 1.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -373,6 +373,9 @@ its \[[Prototype]] an empty object, created as if by ObjectCreate(%ObjectPrototy
 
 <h3 id="logging">Logging methods</h3>
 
+Each {{console}} namespace object has an associated <dfn>count map</dfn>, which is a <a>map</a> of
+<a>strings</a> to labels, initially empty.
+
 <h4 id="assert" oldids="assert-condition-data,dom-console-assert" method for="console">assert(<var ignore=''>condition</var>, ...<var>data</var>)</h4>
 
 <emu-alg>
@@ -398,6 +401,8 @@ Otherwise, do nothing.
 <emu-alg>
   1. Let _called_ be the number of times _count_ has been invoked (including this invocation) with
   the provided _label_.
+    1. If the associated <a>count map</a> contains an entry with key _label_, <a for="map">set</a> the _value_ of map[_label_] to be map[_label_] + 1.
+    1. Else, <a for="map">set</a> the _value_ of the entry with key _label_ in the associated <a>count map</a> to be 1.
   1. Let _concat_ be the concatenation of _label_, U+003A COLON (:), U+0020 SPACE, and
   ToString(_called_).
   1. Perform Logger("count", _concat_).
--- a/index.bs
+++ b/index.bs
@@ -400,7 +400,7 @@ Otherwise, do nothing.
 
 <emu-alg>
   1. Let _map_ be the associated <a>count map</a>.
-  1. If _map_[_label_] <a for="map">exists</a> in <a>map</a>, <a for="map">set</a> _map_[_label_] to _map_[_label_] + 1.
+  1. If _map_[_label_] <a for="map">exists</a>, <a for="map">set</a> _map_[_label_] to _map_[_label_] + 1.
   1. Otherwise, <a for="map">set</a> _map_[_label_] to 1.
   1. Let _concat_ be the concatenation of _label_, U+003A COLON (:), U+0020 SPACE, and
   ToString(_called_).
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -398,6 +401,8 @@ Otherwise, do nothing.
 <emu-alg>
   1. Let _called_ be the number of times _count_ has been invoked (including this invocation) with
```
domfarolino: I might replace this line with the the two you added, since they are a more explicit explanation of what was originally here (this line).

index.bs
```diff
@@ -398,6 +401,8 @@ Otherwise, do nothing.
...
   the provided _label_.
+    1. If the associated <a>count map</a> contains an entry with key _label_, <a for="map">set</a> the _value_ of map[_label_] to be map[_label_] + 1.
+    1. Else, <a for="map">set</a> the _value_ of the entry with key _label_ in the associated <a>count map</a> to be 1.
```
domfarolino: Nit: Else => Otherwise

Probably not super important but I see If/Otherwise used a lot in the spec(s).

index.bs
```diff
@@ -398,6 +401,8 @@ Otherwise, do nothing.
 <emu-alg>
   1. Let _called_ be the number of times _count_ has been invoked (including this invocation) with
   the provided _label_.
+    1. If the associated <a>count map</a> contains an entry with key _label_, <a for="map">set</a> the _value_ of map[_label_] to be map[_label_] + 1.
```
domenic: This doesn't quite work as "map" is an undefined variable. I think this would work better if you had step 1 be "Let _map_ be the associated count map".

You can then also do `_map_[_label_] <a for="map">exists</a>`

musgravejw:
@domfarolino @domenic Thanks for the feedback, I have a commit responding to your suggestions.

index.bs
```diff
@@ -396,8 +399,9 @@ Otherwise, do nothing.
 <h4 id="count" oldids="count-label,dom-console-count" method for="console">count(<var>label</var>)</h4>
 
 <emu-alg>
-  1. Let _called_ be the number of times _count_ has been invoked (including this invocation) with
-  the provided _label_.
+  1. Let map be the associated count map.
```
domenic: Variable names should be italic, so "Let _map_", not "Let map"

index.bs
```diff
@@ -396,8 +399,9 @@ Otherwise, do nothing.
...
+  1. If _map_[_label_] <a for="map">exists</a> in <a>count map</a>, <a for="map">set</a> the _value_ of _map_[_label_] to be _map_[_label_] + 1.
```
domenic: "value" is not a variable, so should not be italic. You can just do "set _map_[_label_] to _map_[_label_] + 1"

index.bs
```diff
@@ -396,8 +399,9 @@ Otherwise, do nothing.
...
+  1. Otherwise, <a for="map">set</a> the _value_ of the entry with key _label_ in the associated <a>count map</a> to be 1.
```
domenic: You can just do "set _map_[_label_] to 1"

index.bs
domenic: There's no need to say "in count map"; _map_ is count map, as you stated in step 1.

musgravejw:
@domenic Hi Domenic, thanks for all the feedback.  Sorry there are a few obvious errors.  Would it be better to reference the existing map in the spec, or define a new one?  I had initially defined a separate map for the label count, which is why I was linking to count map instead of map, but I can take that out if it's unnecessary.

Thanks again, this is my first whatwg PR.

domenic:
> Hi Domenic, thanks for all the feedback. Sorry there are a few obvious errors. Would it be better to reference the existing map in the spec, or define a new one? I had initially defined a separate map for the label count, which is why I was linking to count map instead of map, but I can take that out if it's unnecessary.

Well, you want to define a new one, like you're doing on line 376. But then on line 402, you want to just declare a short alias for it so that you can write your steps in a nice fluid way. It's somewhat similar to

```js
function count(label) {
  const map = this.associatedCountMap;
  
  if (map.has(label)) {
    ...
  }
}
```

instead of

```js
function count(label) {
  if (this.associatedCountMap.has(label)) {
    ...
  }
}
```

> Thanks again, this is my first whatwg PR.

Awesome, happy to have you here!!

Everything looks good to me except one minor linking omission, but I'm happy to fix that up for you.

domfarolino:
@musgravejw We'll want to surround the second argument passed to `Logger()` in the list brackets («*something*») just like we do in [timeEnd](https://console.spec.whatwg.org/#timeend) since Logger takes in a list. Looks like this issue existed before your PR but if you want to fix it that'd be great.

musgravejw:
Cool, thanks @domenic, @terinjokes, @domfarolino !

PR #116, opened 2017-09-02
Meta/Editorial: move to a pure Bikeshed build

domenic:
This removes the emu-algify part of the pipeline. The algorithms and abstract operation cross-links in this spec are not very plentiful, and so they can be handled manually, without needing a second build step.

This includes a few editorial changes as I made a pass through removing <emu-alg> elements:

- Moves "Supporting abstract operations" after "Namespace console"
- Links to more Infra terms, and uses the [=foo/bar=] linking style when appropriate
- Use |var| and `code` syntaxes throughout
- Fixes a couple wrong variables
- By virtue of Bikeshed's default styling, adds spacing in between each algorithm step

For review, I'd suggest previewing the branch snapshot, at https://console.spec.whatwg.org/branch-snapshots/pure-bikeshed/ .

---

/cc @foolip, as I imagine your deploy unification will be easier after this.


foolip:
The emu-algify step could have been supported by https://github.com/whatwg/whatwg.org/pull/81, but if it's not needed, all the better!

PR #117, opened 2017-09-04
Editorial: do not mention emu-algify in README.md

xfq:
I didn't remove the `node_modules/` and `npm-debug.log` entries in `.gitignore`, because they might appear in `reference-implementation/` or `test/`.


domenic:
Thanks!

PR #118, opened 2017-09-04
Meta: use the common deploy.sh (with POST_BUILD_STEP)

foolip:
Part of https://github.com/whatwg/meta/issues/11.


domenic:
A few issues:

- Why doesn't EXTRA_FILES work here?
- Should we add a deploy to the makefile?
- The readme references deploy.sh and so needs an update.

foolip:
> Why doesn't EXTRA_FILES work here?

It's because of the images/ subdirectory. Turns out that `rsync --relative images/*.* deploy/` will create deploy/images/*.*, so I've tweaked this a bit. I could make EXTRA_FILES support this directly, if you think it'll be obvious enough that `EXTRA_FILES="images/*.*"` will create the directory?

> Should we add a deploy to the makefile?

> The readme references deploy.sh and so needs an update.

The POST_BUILD_STEP var needs to be moved to Makefile, but I've made it so, push changes momentarily.

foolip:
This PR now depends on https://github.com/whatwg/whatwg.org/pull/88, branch build will probably fail or be broken.

foolip:
Restarting both builds to see if the world is broken.

foolip:
It worked!

Issue #113, opened 2017-06-05
Handling of Formatter exceptions


terinjokes:
Some of the algorithms defined in the Formatter section can throw exceptions, depending on the input. This specification doesn't state how to handle these exceptions.

```javascript
console.log("symbol: %s", Symbol.for("foo"));
```

In the Formatter algorithm we define how to handle the `%s` formatter:

>If specifier is %s, let converted be the result of ToString(current).

The `ToString` documentation from ECMA-262 states that for Symbol input, a TypeError should be thrown. I would expect, based on the text, that this TypeError is thrown from the Console methods.

Implementations differ in their behavior here, though they do not seem throw the TypeError:
* Chrome 59 prints: `symbol: Symbol(foo)`
* Safari 10 prints: `symbol: Symbol(foo)`
* Firefox 54 prints:
* (TODO: Edge)

domfarolino:
This seemed closely related to https://github.com/whatwg/console/issues/26 and https://github.com/whatwg/console/issues/88 regarding `ToString` conversion stuff. Therefore I felt like implementations might start throwing here just as a result of fixing some of the previous issues (like re-throwing other `ToString` conversions https://bugzilla.mozilla.org/show_bug.cgi?id=1346336), however latest FF does not throw in your example (or any `ToString` failure with the `%s` specifier).

Given the decisions we've previously made (to re throw other `ToString` errors) I think we should add spec text to re-throw the error even though no implementations do that now. If agreed, I can add a PR for this and file some more browser issues.

domenic:
Although I agree with the general principle, I think in the particular case here, throwing for all symbols would be bad. Instead we might want to special case them.

JosephPecoraro:
> Given the decisions we've previously made (to re throw other ToString errors) I think we should add spec text to re-throw the error even though no implementations do that now.

Can you point me to that discussion?

---

WebKit (and Blink as I understand it) don't currently evaluate the format string ("Formatter") operations in the page's context. Instead each argument is captured and serialized to a debugger. The debugger then applies the format string operations as needed.

One advantage to this approach is that console operations when there is no debugger opened can do very little work (just hold on to the values). Another advantage is that there are no exceptions using `console.log`, so developer feel free to sprinkle them around and not break program control flow.

Also, changing `console.log` behavior to now throw exceptions has the potential to break existing web pages with log statements that wouldn't have thrown before. Is that a concern?

domfarolino:
> Can you point me to that discussion?

[...]

PR #123, opened 2017-12-21
Fix Formatter to use String() instead of ToString()

domfarolino:
First attempt at #113. I see two ways to go about this:

 - [Initial attempt] Use `String()` instead `ToString()` when Formatter converts arguments with the `%s` specifier. Behavior is the same as `ToString()` except it doesn't throw when converting Symbols. Not sure if this is a viable since `String()` isn't an abstract internal-only operating (are we limited to using these at all? Is it bag to call a constructor in this kinda position?)
 - Run a `Type()` check before we convert with `ToString()`, and call [`SymbolDescriptiveString`](https://tc39.github.io/ecma262/#sec-symboldescriptivestring) in the case we have a Symbol. (Basically performing the logic of the `String()` constructor ourselves).

Both of the above ways will prevent Formatter from throwing so the changes aren't big, and are pretty narrow in scope.

### Edit

I'll add some wpt for this as well once it is approved.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -35,6 +35,8 @@ urlPrefix: https://tc39.github.io/ecma262/#; spec: ECMASCRIPT
     text: %parseInt%; url: sec-parseint-string-radix
   type: interface
     text: %ObjectPrototype%; url: sec-properties-of-the-object-prototype-object
+  type: constructor
+    text: String; url: sec-string-constructor
 </pre>
 
 <style>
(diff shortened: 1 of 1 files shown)


domenic:
Hmm. So one issue here is that you need to use `%String%` instead of just `String` to ensure you're using the built-in constructor.

Another issue is a preexisting issue, which is that you indeed shouldn't be treating %String%, or %parseInt%, or %parseFloat%, as abstract ops; you need to instead do e.g. `Call(%parseInt%, undefined, « |current|, 10 »)`.

All that said: at first I was leaning toward your second option, of inlining the logic. But given that the next few lines need to do very similar things for reusing the parseInt/parseFloat logic, I think your version (but with the %s inserted) is the way to go. Then, as a separate thing, we can clean up the abstract ops vs. functions thing later.

domfarolino:
Chrome and Safari seem up-to-spec with this. Here's the Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1429971

@domenic would you mind adding Edge's result?

PR #130, opened 2018-03-27
 chore: replace babel-preset-es2015 with babel-preset-env 

hiroppy:
And update README.

### DIFF
--- a/reference-implementation/.babelrc
+++ b/reference-implementation/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["es2015"]
+  "presets": ["env"]
 }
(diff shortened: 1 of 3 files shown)


hiroppy:
Thank you for review:)

domfarolino:
I'm not really sure how effective it will be to update the reference implementation here. I know it doesn't get too much TLC these days; I'm not opposed to getting back on top of it, but it also doesn't seem terribly necessary. I think once you're verified (cc @annevk ) and we have a decision on that, we can move forward with this or request some changes. Thanks!

annevk:
Verified!

hiroppy:
@annevk Thanks!

domenic:
I think removing the reference implementation would be a good idea, personally, given how it hasn't kept pace and how the tests should be redundant with web-platform-tests these days.

PR #131, opened 2018-03-28
Editorial: update NOTES.md

hiroppy:
Update the version of Chrome to 65.0.3325.181.


hiroppy:
I think it would be better to specify the browser version clearly:)

domfarolino:
Thanks for the PR. A couple things here:

 - I'm honestly not sure how useful the `NOTES.md` file is anymore. I imagine it was really helpful in the early stages of the spec to help define what is important/missing etc., but I know we haven't really seemed to update it regularly (e.g., latest findings on format specifiers + Symbols).
 - Piggybacking off of the previous point, in general I'm against consistently keeping the file up-to-date with what each implementation (and their version #s) is doing in the near-latest release. Things are always changing, and it's a battle that can't be won; also once the browser issues are filed and [WPTs](https://github.com/w3c/web-platform-tests) are written, our job is basically done (though making sure browsers follow through on implementations is good to do too).

`console.table` seems to be a good counterexample since it's algorithm is not-yet spec'd, but always keeping up with how it looks in each latest browser isn't really fun. It *would* be nice if we could always see the behavior of each up-to-date Console implementation, but ultimately we can't really keep track of that all the time so I think we should loop in @terinjokes and maybe Domenic (when back from vacation) to see if it is worth deprecating/deleting the Notes file altogether.

hiroppy:
Yeah, I agree with @domfarolino.
I think it is very difficult to always keep NOTES.md up to date.(If it is to automate it may be different 😉)

domfarolino:
Terribly sorry to keep this waiting, I have a few ideas for this, so let me circle back here in a bit. Thank you!

domfarolino:
I just filed https://github.com/whatwg/console/issues/181 to track removing `NOTES.md`, so I'm actually going to close this, since we're planning on ripping it out. Thanks for the PR and working to keep this repository up-to-date!

PR #137, opened 2018-05-21
Fix leftover ToString reference in format specifier table's %s entry

fmartin5:
Hello, I could find some time to work on this, though I am not experienced with Bikeshed.

There is another `%String%` reference in the spec, generated using the `{{%String%}}` syntax; however it appears as an orangered link, not a blue one like for the `%parseInt%` and `%parseFloat%` functions.

Is this behavior intended so I should have used the same syntax, or is there something to do about both references?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -314,7 +314,7 @@ The following is an informative summary of the format specifiers processed by th
   <tr>
     <td>`%s`</td>
     <td>Element which substitutes is converted to a string</td>
-    <td><a constructor>%String%</a>(|element|)</td>
+    <td><a idl>%String%</a>(|element|)</td>
   </tr>
   <tr>
     <td>`%d` or `%i`</td>
--- a/index.bs
+++ b/index.bs
@@ -277,7 +277,7 @@ more arguments are left. It returns a [=list=] of objects suitable for printing.
 1. Let |current| be the second element of |args|.
 1. Find the first possible format specifier |specifier|, from the left to the right in |target|.
   1. If |specifier| is `%s`, let |converted| be the result of
-     [$Call$]({{%String%}}, **undefined**, « |current| »).
+     [$Call$](<a idl>%String%</a>, **undefined**, « |current| »).
   1. If |specifier| is `%d` or `%i`:
     1. If [$Type$](|current|) is Symbol, let |converted| be `NaN`
     1. Otherwise, let |converted| be the result of
--- a/index.bs
+++ b/index.bs
@@ -314,7 +314,7 @@ The following is an informative summary of the format specifiers processed by th
   <tr>
     <td>`%s`</td>
     <td>Element which substitutes is converted to a string</td>
-    <td><a abstract-op>ToString</a>(|element|)</td>
+    <td><a constructor>%String%</a>(|element|)</td>
   </tr>
   <tr>
     <td>`%d` or `%i`</td>
--- a/index.bs
+++ b/index.bs
@@ -523,6 +523,7 @@ Ian Kilpatrick,
 Jeff Carpenter,
 Joseph Pecoraro,
 Justin Woo,
+Luc Martin,
 Noah Bass,
 Paul Irish,
 Raphaël, and


domfarolino:
Some of the bikeshed documentation might helpful here to understand what is going on (https://tabatkins.github.io/bikeshed/#dfn-types and https://tabatkins.github.io/bikeshed/#autolink-wrappers specifically). These touch a bit on the different definition-wrapping syntaxes.

Having the two references of `%String%` in the spec differ in style is not intended, and we can probably do something about that. With `%String%`, we're referencing the %String% internal constructor defined in the ECMAScript spec. You can see that we're letting Bikeshed know it is a constructor in the [beginning of the spec source](https://github.com/whatwg/console/blob/master/index.bs#L26). When we want to link to it, we can do so in several ways:

 - `<a constructor>%String%</a>` (results in a normal blue link)
 - `<a idl>%String%</a>` (same as above; `idl` serves as a union to any of the WebIDL types (interface, constructor, etc., so it's good to use when you don't *have* to specify)
 - `{{%String%}}` (orange code-style link) (short for `<a idl><code>%String%</code></a>`, though the style is a bit different it seems, given the light underline in the former).

The orange-style link seems to be used mostly for exposed APIs/objects; for example, see [`Request()`](https://fetch.spec.whatwg.org/#dom-request), which is the definition of the [`[Constructor...]`](https://fetch.spec.whatwg.org/#request-class) defined in the second line of this IDL file.

Again, we're referring to the internal %String% constructor defined by ECMAScript, which isn't exposed (of course there is an exposed `String()` constructor). So for this we should probably be using `<a idl>%String%</a>` anywhere we can. I'd also confirm my thoughts with @domenic to make sure I understand when & where we'd use what.

index.bs
```diff
@@ -314,7 +314,7 @@ The following is an informative summary of the format specifiers processed by th
   <tr>
     <td>`%s`</td>
     <td>Element which substitutes is converted to a string</td>
-    <td><a abstract-op>ToString</a>(|element|)</td>
+    <td><a constructor>%String%</a>(|element|)</td>
```
domfarolino: It looks like we can get rid of the [ToString](https://github.com/whatwg/console/blob/master/index.bs#L22) URL prefix now that we have no more references to it, if you'd like to remove that as well. Also feel free to put your name in the acknowledgements.

domfarolino:
Thanks a lot! Just waiting for the removal of the ToString [anchor reference](https://github.com/whatwg/console/blob/master/index.bs#L22) and we should be good to go.

(Also if you wanted to fix the other instance of `{{%String%}}` => `<a idl>%String%</a>` that certainly wouldn't go amiss :) )

fmartin5:
Thank you for the detailed comments and explanations!

Regarding the `ToString` URL prefix, there is still a reference to it in the `console.count` method
at [line 117](https://github.com/whatwg/console/blob/master/index.bs#L117):

> 4. Let concat be the concatenation of label, U+003A (:), U+0020 SPACE, and ToString(map[label]).

domenic:
I don't have strong feelings on the choice between `<a idl>%String%</a>` and `{{%String%}}`. I guess I lean toward the latter since it's what the Bikeshed shorthands guide you toward, but *shrug*.

domfarolino:
Good catch on the other ToString. I’m good with this PR unless you’d like use the bikeshed shorthand notation, else I’ll merge soon (currently on mobile so I’ll take another look on desktop soon).

domfarolino:
Also do you want to add yourself to the acknowledgments section (bottom of spec)?

fmartin5:
I would also prefer the more idiomatic wrapper syntax over manual autolinks, if it was not for these points:
- other entries in the format specifier table already use them;
- the `{{%String%}}` syntax produces a link which is visually different from other EcmaScript definition links in the same table, albeit visually similar to the `%ObjectPrototype%` links.

Do you have thoughts about the validity of these points and which tradeoff is better, or is there a way to use the wrapper syntax without these issues?

For example, is there a possibility to move the %String% anchor reference from the `type: constructor` section to the `type: dfn` section?
- this would allow the `[=%String%=]` syntax to display ordinary blue dfn links like it currently does for %parseFloat% and %parseInt%.

- %String% is an ECMAScript constructor,
 and I am confused as to why it is also regarded as a WebIDL constructor in the console spec by using `type: constructor` and orange links, since the ECMAScript spec does not contain any WebIDL code.
- however using `type: dfn` causes lowercasing in the "Terms defined by reference" section, like it already does for %parseFloat% and %parseInt%.

Regarding the acknowledgements, should I put my real name or the name figuring on my GithHub account?

I am asking this because some people visibly did not put their real full name, e.g. [Raphaël](https://github.com/whatwg/console/blob/master/index.bs#L528), or [Ms2ger](https://github.com/heycam/webidl/blob/master/index.bs#L13496) in the WebIDL spec.

domfarolino:
*Disclaimer: a lot of my points are based on the Streams Standard source*

> albeit visually similar to the {{%ObjectPrototype%}} links

This is another scenario I've thought about changing to the `<a idl>...</a>` form, mostly because of how the same link appears in Streams.

> Do you have thoughts about the validity of these points and which tradeoff is better

I prefer the long-hand syntax for things like internal definitions and abstract operations, and the shorthand syntax for things that are exposed via JS as something directly usable by application code. For example:

 - See "DataView" in https://streams.spec.whatwg.org/#readable-byte-stream-controller-pull-into
 - Worker() [here](https://html.spec.whatwg.org/multipage/workers.html#dom-worker) is an exposed API, and the orange/code-style syntax generally makes that clear to me, whereas the use of ["run a worker"](https://html.spec.whatwg.org/multipage/workers.html#run-a-worker) is not, and is an abstract operation so it is simply **bolded** in definition, and has regular style blue links where its used.

I believe I've asked Tab before and there is no way to use the shorthand wrapper syntax without getting the orange/code-style links.

Regarding the other points, personally I'd keep `%String%` in the `constructor` section, but I'm basing judgement here on %Uint8Array% and 
Uint8Array in the Streams Standard (see https://github.com/whatwg/streams/blob/master/index.bs#L20), whose usages and accompanying syntax I jive with. In other words, I'm not sure how strict that "WebIDL" types "restriction" is here, but as I mentioned I prefer to use the IDL types for a union of IDL types and exposed things.

> should I put my real name or...

Up to you; we have a lot of "real names" in this spec so it might look cleaner, but doesn't matter, as long as the list remains alphabetical.

domfarolino:
Honestly doesn't matter too much @fmartin5, but do we have your name wrong [here](https://github.com/whatwg/participant-data/commit/7daaa093686e0266cc1d0a526bff290336baf0a8#diff-72ba56fd36e9601732bcb2c90e594ba6R609)?

fmartin5:
No, that was my real name in the participant data, whereas Luc Martin is the name I use when freelancing and on my GitHub account, as well as on a few other online communities.

Issue #134, opened 2018-05-04
time() should call Logger(); timeEnd should report warning


domfarolino:
Right now `console.time()` mentions an optionally-reported warning on trying to start an already-existing timer. We should strengthen this by calling `Logger()` when necessary.

Furthermore, `console.timeEnd()` in all implementations (can't remember Edge's behavior though) print warnings when trying to end a non-existent timer. To bring the spec closer to implementations, and also align `timeEnd()` with `countReset` (currently being specced), I propose we add language making `timeEnd()` report an error under the previously-mentioned conditions. I also propose that we move the `timeEnd()` entry in the logLevel-severity table to the `log` category, to guide implementations a bit in the future, since nothing I've tested has visible UX difference (anymore, Chrome used to at least) betwee `timeEnd()` and `log()`.

PR #138, opened 2018-05-25
Specify console.timeLog() + clean up timing

### LINKED ISSUE #84: Intermediate values with console.time
There should be a way to print how much time has passed since a timer has started without removing the timer.

In Chrome calling `timeEnd` currently doesn't remove the timer, so you can call timeEnd multiple times.

    console.time("abc")
    console.timeEnd("abc")
    console.timeEnd("abc")
    // abc: 0.006ms
    // abc: 0.281ms

The timer is reset only when `time` is called again with the same label. Firefox and Node remove the timer when `timeEnd` is called, as the spec says.

To be able to see intermediate values the spec could either be changed to recommend Chrome's behavior, or a new `timeLog` (or similar) function could be introduced, which prints the duration without affecting the timer itself.

    console.time("abc")
    console.timeLog("abc")
    console.timeEnd("abc")

domfarolino:
This PR:

 - Adds `timeLog` to the Console Standard (and specifies that it should use `Printer` instead of `Logger` so we ignore format specifiers appearing in a timer label.
 - Cleans up the way we interact with associated `timer table`s in the timing methods - this is a little prep work for later when we check (non)existence of `timer table` entries so we can formally report warnings.
 - Points to an issue #134 to show plans for having the timing methods formally report warnings.
 - Fixes `timeEnd` to actually remove a timer from the timer table (the reason we went with adding `timeLog`)
 - Adds an example show how someone might use the extra `...data` passed into `timeLog` to distinguish calls to `timeLog` from each other throughout the duration of a timer.
 - Fixes `countReset`'s permalink to be all lowercase (that seems to be the trend here)
 - Closes #84 

Interested in getting some opinions from @xirzec, @nchevobbe, and @JosephPecoraro on this to make sure we have consensus and can get this in ASAP! (will file browser bugs soon)

Bugs
 - Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1461001
 - Safari: https://bugs.webkit.org/show_bug.cgi?id=186833
 - Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=854474
 - Edge: https://twitter.com/domfarolino/status/1009303140556931073

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -68,8 +68,8 @@ namespace console { // but see namespace object requirements below
 
   // Timing
   void time(optional DOMString label = "default");
-  void timeEnd(optional DOMString label = "default");
   void timeLog(optional DOMString label = "default", any... data);
+  void timeEnd(optional DOMString label = "default");
 };
 </pre>
 
@@ -234,18 +234,6 @@ Each {{console}} namespace object has an associated <dfn>timer table</dfn>, whic
 1. Otherwise, [=map/set=] the value of the entry with key |label| in the associated
    <a>timer table</a> to the current time.
 
-<h4 id="timeend" oldids="timeend-label,dom-console-timeend" method for="console">timeEnd(|label|)</h4>
-
-1. Let |timerTable| be the associated <a>timer table</a>.
-1. Let |startTime| be |timerTable|[|label|].
-1. [=map/set|Remove=] |timerTable|[|label|].
-1. Let |duration| be a string representing the difference between the current time and
-   |startTime|, in an implementation-defined format.
-   <p class="example" id="duration-string-example">"4650", "4650.69 ms", "5 seconds", and "00:05"
-   are all reasonable ways of displaying a 4650.69 ms duration.</p>
-1. Let |concat| be the concatenation of |label|, U+003A (:), U+0020 SPACE, and |duration|.
-1. Perform <a abstract-op>Printer</a>("timeEnd", « |concat| »).
-
 <h4 id="timelog" method for="console">timeLog(|label|, ...|data|)</h4>
 
 1. Let |timerTable| be the associated <a>timer table</a>.
@@ -273,6 +261,18 @@ Each {{console}} namespace object has an associated <dfn>timer table</dfn>, whic
     console.timeEnd("MyTimer");
   </code></pre>
 </div>
+
+<h4 id="timeend" oldids="timeend-label,dom-console-timeend" method for="console">timeEnd(|label|)</h4>
+
+1. Let |timerTable| be the associated <a>timer table</a>.
+1. Let |startTime| be |timerTable|[|label|].
+1. [=map/Remove=] |timerTable|[|label|].
+1. Let |duration| be a string representing the difference between the current time and
+   |startTime|, in an implementation-defined format.
+   <p class="example" id="duration-string-example">"4650", "4650.69 ms", "5 seconds", and "00:05"
+   are all reasonable ways of displaying a 4650.69 ms duration.</p>
+1. Let |concat| be the concatenation of |label|, U+003A (:), U+0020 SPACE, and |duration|.
+1. Perform <a abstract-op>Printer</a>("timeEnd", « |concat| »).
 
 <p class="note">See <a href="https://github.com/whatwg/console/issues/134">whatwg/console#134</a>
 for plans to make {{console/timeEnd()}} and {{console/timeLog()}} formally report warnings to the
@@ -428,7 +428,7 @@ their output similarly, in four broad categories. This table summarizes these co
     <td>log</td>
     <td>
       {{console/log()}}, {{console/trace()}}, {{console/dir()}}, {{console/dirxml()}},
-      {{console/group()}}, {{console/groupCollapsed()}}, {{console/debug()}} {{console/timeLog()}}
+      {{console/group()}}, {{console/groupCollapsed()}}, {{console/debug()}}, {{console/timeLog()}}
     </td>
     <td>
       A generic log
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -231,14 +231,48 @@ Each {{console}} namespace object has an associated <dfn>timer table</dfn>, whic
...
+    // Perhaps some fetch()'s here filling the app with data
+    // ...
+    console.timeLog("MyTimer", "All done!");
+    console.timeEnd("MyTimer");
+  </code></pre>
+</div>
+
+<p class="note">See <a href="https://github.com/whatwg/console/issues/134">whatwg/console#134</a>
+for plans to make {{console/timeEnd()}} and {{console/timeLog()}} formally report warnings to the
+console when a give |label| does not exist in the associated <a>timer table</a>.
```
fmartin5: A small typo here, the word "given" is misspelled. This looks good otherwise :)

index.bs
```diff
@@ -231,14 +231,48 @@ Each {{console}} namespace object has an associated <dfn>timer table</dfn>, whic
 
 <h4 id="timeend" oldids="timeend-label,dom-console-timeend" method for="console">timeEnd(|label|)</h4>
 
-1. Let |startTime| be the result of [=map/getting=] the value of the entry with key |label| in the
-   associated <a>timer table</a>.
+1. Let |timerTable| be the associated <a>timer table</a>.
+1. Let |startTime| be |timerTable|[|label|].
+1. [=map/set|Remove=] |timerTable|[|label|].
```
domfarolino: This should probably `map/remove` but I have to take a closer look

domenic: You should be able to just do `[=map/Remove=]`

index.bs
```diff
@@ -389,7 +423,7 @@ their output similarly, in four broad categories. This table summarizes these co
     <td>log</td>
     <td>
       {{console/log()}}, {{console/trace()}}, {{console/dir()}}, {{console/dirxml()}},
-      {{console/group()}}, {{console/groupCollapsed()}}, {{console/debug()}}
+      {{console/group()}}, {{console/groupCollapsed()}}, {{console/debug()}} {{console/timeLog()}}
```
domenic: Missing comma

index.bs
```diff
@@ -231,14 +231,48 @@ Each {{console}} namespace object has an associated <dfn>timer table</dfn>, whic
...
+1. [=map/set|Remove=] |timerTable|[|label|].
 1. Let |duration| be a string representing the difference between the current time and
    |startTime|, in an implementation-defined format.
    <p class="example" id="duration-string-example">"4650", "4650.69 ms", "5 seconds", and "00:05"
    are all reasonable ways of displaying a 4650.69 ms duration.</p>
 1. Let |concat| be the concatenation of |label|, U+003A (:), U+0020 SPACE, and |duration|.
-1. Perform <a abstract-op>Logger</a>("timeEnd", « |concat| »).
+1. Perform <a abstract-op>Printer</a>("timeEnd", « |concat| »).
+
+<h4 id="timelog" method for="console">timeLog(|label|, ...|data|)</h4>
```
domenic: I wonder if putting "log" between `time` ("start") and "end" might make be nicer?

domfarolino:
Rebased this as well. Will get moving on bugs + tests.

index.bs
```diff
@@ -234,16 +234,50 @@ Each {{console}} namespace object has an associated <dfn>timer table</dfn>, whic
 1. Otherwise, [=map/set=] the value of the entry with key |label| in the associated
    <a>timer table</a> to the current time.
 
+<h4 id="timelog" method for="console">timeLog(|label|, ...|data|)</h4>
+
+1. Let |timerTable| be the associated <a>timer table</a>.
+1. Let |startTime| be |timerTable|[|label|].
+1. Let |duration| be a string representing the difference between the current time and
+   |startTime|, in an implementation-defined format.
+1. Let |concat| be the concatenation of |label|, U+003A (:), U+0020 SPACE, and |duration|.
```
nchevobbe: should we indicate what the duration represent (milliseconds, seconds, ...) ?
It seems that there is a bit more information about it in the timeEnd section.

index.bs
```diff
@@ -234,16 +234,50 @@ Each {{console}} namespace object has an associated <dfn>timer table</dfn>, whic
...
+1. Perform <a abstract-op>Printer</a>("timeLog", data).
+
+<div class="example" id="timelog-example">
+  The |data| parameter in calls to {{console/timeLog()}} is included in the call to
+  <a abstract-op>Logger</a> to make it easier for users to supply intermediate timer logs with
+  some extra data throughout the life of a timer. For example:
+
+  <pre><code class="lang-javascript">
+    console.time("MyTimer");
+    console.timeLog("MyTimer", "Starting application up...");
```
nchevobbe: nit: Replace `...` with the proper ellipsis char `…`

index.bs
```diff
@@ -234,16 +234,50 @@ Each {{console}} namespace object has an associated <dfn>timer table</dfn>, whic
...
+    // Perhaps some code runs to boostrap a complex app
+    // ...
+    console.timeLog("MyTimer", "UI is setup, making API calls now");
+    // Perhaps some fetch()'s here filling the app with data
+    // ...
+    console.timeLog("MyTimer", "All done!");
```
nchevobbe: Not sure if this case would make sense since `timeEnd`already prints the final timer duration.

index.bs
domfarolino: I think what I'll do here is move the example in timeEnd up to here since the first instance of time represented in an implementation-defined format should probably do the explaining. Probably don't need to add it again in timeEnd since it's a repeat step and already will have an example here.

xirzec:
Also seems reasonable to me. /cc @thelarkinn

PR #145, opened 2018-06-21
Editorial: fix countReset permalink

domfarolino:
To match the style of other permalinks. There are a few links to the camelCase version of think link in WPTs which I'll update separately.


domfarolino:
Re-kicked the build to ensure that it passed, since last time it seems api.csswg.org was down.

PR #151, opened 2018-11-24
Report warning

### LINKED ISSUE #57: Formalize warnings
Often specifications want to give hints at things that can be reported to a developer console, if there is any. Should we formalize that in the Console Standard?

(I was looking at fixing https://github.com/whatwg/fetch/issues/259 and then I wondered whether we should have some shared mechanism across specifications.)

domfarolino:
Closes #57. Not entirely familiar with creating an exportable definition in Bikeshed, but given some examples in whatwg/dom and friends I take it the `export` keyword is sufficient for Bikeshed to pick it up?

Also I'm happy to make a `ReportWarning(...)` abstract operation instead of just a dfn that can be given "a generic warning", but I figured I'd go down this route first.

Edit: #57 mentions it might be good to make this _optionally call Printer(...)_, but personally I prefer to make it mandatory, and have callers _optionally_ call into our definition. I think this will make their specs more clear.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -550,6 +550,13 @@ representation of an object judged to be maximally useful and informative.
   </code></pre>
 </div>
 
+<h3 id="reporting-warnings">Reporting warnings to the console</h3>
+
+This specification defines the concept of <dfn export lt="report a warning">reporting a
+warning</dfn>, which can be used by other specifications when it is desireable to display a warning
+in the developer console. To <a>report a warning</a> given a generic implementation-defined warning,
+implementations must perform <a abstract-op>Printer</a>("reportWarning", « |warning| », |options|).
+
 <h2 id="acks" class="no-num">Acknowledgments</h2>
 
 The editors would like to thank
--- a/index.bs
+++ b/index.bs
@@ -553,7 +553,7 @@ representation of an object judged to be maximally useful and informative.
 <h3 id="reporting-warnings">Reporting warnings to the console</h3>
 
 To <dfn export>report a warning to the console</dfn> given a generic description of a warning
-|description|, implementations must perform:
+|description|, implementations must run these steps:
 
 1. Let |warning| be an implementation-defined string derived from |description|.
 1. Perform <a abstract-op>Printer</a>("reportWarning", « |warning| »).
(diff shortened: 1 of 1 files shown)


domenic:
Overall looks good. Might be able to simplify it; the structure of "This specification defines X... to X, perform the following steps..." is a bit strange. But no immediate better idea.

But, one thing I think should be addressed is to explain how we don't generally specify exact strings. The proposed algorithm gives the impression that you would pass a string.

You could fix this a couple of ways. One would be to explain the above, and maybe add an example of an algorithm that calls this, which says something like

> If the _foo_ value is "`bar`", <a>report a warning to the console</a> saying that "`bar`" has no effect, and return.

This formulation is basically saying that callers should feed descriptions of _message_ to the algorithm, and not exact strings.

Alternately, you could make this pattern more explicit in the definition of the algorithm, e.g. saying "given a description of a message _description_", and calling Printer with "some implementation-defined string derived from _description_".

domfarolino:
> the structure of "This specification defines X... to X, perform the following steps..." is a bit strange. But no immediate better idea.

Yeah that makes sense, my main goal was to point out that basically this is something that's "meant" to be consumed by specs other than Console (and is exported), however that is probably obvious from context; for now I removed it.

Ping for re-review!

index.bs
```diff
@@ -550,6 +550,14 @@ representation of an object judged to be maximally useful and informative.
   </code></pre>
 </div>
 
+<h3 id="reporting-warnings">Reporting warnings to the console</h3>
+
+To <dfn export>report a warning to the console</dfn> given a generic description of a warning
+|description|, implementations must perform:
```
fmartin5: Nit: for stylistic consistency with other specs I would favor the construct
`implementations must run these steps: ...` over
 `implementations must perform: ...`

For example, in the DOM standard at https://dom.spec.whatwg.org the text "must run these steps" currently occurs 73 times wheras "must perform" does not seem to appear at all.

domfarolino:
I was actually going to do all of that separately as a part of https://github.com/whatwg/console/issues/134, so stay tuned!

PR #162, opened 2019-07-08
Editorial: methods -> functions

### LINKED ISSUE #161: Console operations are functions
Discussed in https://github.com/WICG/badging/pull/38#issuecomment-508628566 with @domenic as to whether namespaces have "methods" or "functions". Conclusion was that they are functions, because they don't operate on `this`.

marcoscaceres:
fixes #161


marcoscaceres:
>  a little confusing with the method attribute, but that's probably alright if this is the preferred way to go.

Yeah, I spotted that too... I assume those are export annotations for BikeShed. Should be fine if so.

domfarolino:
Yeah they are, and removing them breaks links it seems. Merging

Issue #159, opened 2019-06-26
Make first argument of `console.dir` and `console.table` optional


dcrousso:
WebKit, Chromium, and Gecko all support calling `console.dir` and `console.table` with no first argument (e.g. no exception is thrown), so the spec should match these implementations.  Given that every other `console` method allows for any number of arguments without throwing an error, it's reasonable to have `console.dir` and `console.table` do the same.

The implications of this is that `console.dir.length` and `console.table.length` would now be equal to `0`. 

```diff
 [Exposed=(Window,Worker,Worklet)]
 namespace console { // but see namespace object requirements below
   // Logging
   void assert(optional boolean condition = false, any... data);
   void clear();
   void debug(any... data);
   void error(any... data);
   void info(any... data);
   void log(any... data);
-  void table(any tabularData, optional sequence&lt;DOMString> properties);
+  void table(optional any tabularData, optional sequence&lt;DOMString> properties);
   void trace(any... data);
   void warn(any... data);
-  void dir(any item, optional object? options);
+  void dir(optional any item, optional object? options);
   void dirxml(any... data);
 
   // Counting
   void count(optional DOMString label = "default");
   void countReset(optional DOMString label = "default");
 
   // Grouping
   void group(any... data);
   void groupCollapsed(any... data);
   void groupEnd();
 
   // Timing
   void time(optional DOMString label = "default");
   void timeLog(optional DOMString label = "default", any... data);
   void timeEnd(optional DOMString label = "default");
 };
```

NOTE: this is why WebKit (and Gecko) is failing the [idlharness.any](https://wpt.fyi/results/console/idlharness.any.html) and [idlharness.worker.any](https://wpt.fyi/results/console/idlharness.any.worker.html) tests.

JosephPecoraro:
What is the next step on this? Can we get it reviewed / integrated?

domenic:
I see there's a pull request at https://github.com/whatwg/console/pull/164; hopefully one of the editors will review it soon. I guess the web platform tests will be automatically updated by the IDL-scraping tools?

JosephPecoraro:
> I guess the web platform tests will be automatically updated by the IDL-scraping tools?

That was my understanding.

domfarolino:
Closes with the above-referenced #164

PR #164, opened 2019-07-19
Make first argument of `console.dir` and `console.table` optional (#159)

dcrousso:
WebKit, Chromium, and Gecko all support calling `console.dir` and `console.table` with no first argument (e.g. no exception is thrown), so the spec should match these implementations.  Given that every other `console` method allows for any number of arguments without throwing an error, it's reasonable to have `console.dir` and `console.table` do the same.

The implications of this is that `console.dir.length` and `console.table.length` would now be equal to `0`. 

```diff
 [Exposed=(Window,Worker,Worklet)]
 namespace console { // but see namespace object requirements below
   // Logging
   void assert(optional boolean condition = false, any... data);
   void clear();
   void debug(any... data);
   void error(any... data);
   void info(any... data);
   void log(any... data);
-  void table(any tabularData, optional sequence&lt;DOMString> properties);
+  void table(optional any tabularData, optional sequence&lt;DOMString> properties);
   void trace(any... data);
   void warn(any... data);
-  void dir(any item, optional object? options);
+  void dir(optional any item, optional object? options);
   void dirxml(any... data);
 
   // Counting
   void count(optional DOMString label = "default");
   void countReset(optional DOMString label = "default");
 
   // Grouping
   void group(any... data);
   void groupCollapsed(any... data);
   void groupEnd();
 
   // Timing
   void time(optional DOMString label = "default");
   void timeLog(optional DOMString label = "default", any... data);
   void timeEnd(optional DOMString label = "default");
 };
```

NOTE: this is why WebKit (and Gecko) is failing the [idlharness.any](https://wpt.fyi/results/console/idlharness.any.html) and [idlharness.worker.any](https://wpt.fyi/results/console/idlharness.any.worker.html) tests.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -51,10 +51,10 @@ namespace console { // but see namespace object requirements below
   void error(any... data);
   void info(any... data);
   void log(any... data);
-  void table(any tabularData, optional sequence&lt;DOMString> properties);
+  void table(optional any tabularData, optional sequence&lt;DOMString> properties);
   void trace(any... data);
   void warn(any... data);
-  void dir(any item, optional object? options);
+  void dir(optional any item, optional object? options);
   void dirxml(any... data);
 
   // Counting


JosephPecoraro:
Thanks! Should this close https://github.com/whatwg/console/issues/159?

domfarolino:
Oh interesting. I was thinking my “Closes #159” above would do that automatically. Yeah I’ll close.

PR #169, opened 2020-04-01
Meta: update repository files

annevk:
See https://github.com/whatwg/spec-factory for details.


annevk:
See https://github.com/whatwg/spec-factory/issues/1 for some discussion on the `.editorconfig` changes.

.editorconfig
```diff
@@ -4,10 +4,13 @@ root = true
 end_of_line = lf
 insert_final_newline = true
 charset = utf-8
-indent_size = 2
+indent_size = 1
```
terinjokes, line 7: Does this change cause any changes in the actual spec? If so, should we be including it as part of this change?

annevk: That's a good question. Happy to leave that out this round, but I do think it would be good for Console to align on this.

terinjokes: Sorry, I meant should we be including the changes _to the spec_ as part of this. I don't use editorconfig, so it will take me a bit to find out myself, and it's clear I should already be sleeping. ;)

annevk: If someone is willing to reformat the specification we could include that change here I suppose, but it might be better as a distinct commit.

domenic:
I tend to agree that we should keep the editorconfig and the spec itself in sync.

.editorconfig
annevk, line 7:
```suggestion
indent_size = 2
```

PR #180, opened 2020-06-15
Meta: Remove the reference implementation

### LINKED ISSUE #152: Address security warning on the reference implementation
lodash needs to be at least 4.17.5 per GitHub alerts.

domfarolino:
This PR removes the reference implementation and the tests associated with it. The reference implementation for this standard is fairly un-maintained and hasn't been updated in a while. My understanding is that no implementations really use it as a reference to see how their console interface should act or look, so I think it is reasonable to remove this from the repository.

This will also fix #152, which is about updating one of the reference implementation dependencies.


annevk:
This might also allow for some changes to https://github.com/whatwg/spec-factory/blob/master/factory.json.

domenic:
Perhaps remove the `test/` directory as well.

domfarolino:
> Perhaps remove the test/ directory as well.

I was thinking about this too, but figured we might want to do it separately, since there could be some tests we can port to WPTs

PR #205, opened 2021-12-13
Handle input with no format specifiers in `Formatter`

### LINKED ISSUE #204: Formatter algorithm does not handle having no format specifiers
https://console.spec.whatwg.org/commit-snapshots/6447e9d3cdb01c48094cbaa32d0e5fb1ab74367d/#formatter

Hello! I'm in the process of implementing the console spec in a new browser, and came across some issues with using `result` in the *Formatter* algorithm.

Step 4 in the *Formatter* algorithm asks us to return `result`:
> 4. If target does not have any format specifiers left, return result.

If the first element of `args` did not contain any format specifiers, then `result` has no value here. This can occur, since `trace()`, `group()` and `groupCollapsed()` all call *Formatter* without checking that a format specifier is present.

I think this is a typo, and step 3.8 should be unindented and appear as step 4. However, if no format specifier was found, that would skip the second element of `args`. So either this needs special handling, or the functions calling *Formatter* all need to ensure that it does contain a format specifier first, and call *Printer* otherwise.

AtkinsSJ:
Hi! This is my first contribution to a spec like this so while I've done my best to follow the guidelines, there are probably things I am getting wrong. Let me know and I'll try and sort them out as soon as possible.

---

This fixes #204.

As noted in that issue, several functions call `Formatter` without
checking that a format specifier is present, but the `Formatter`
algorithm assumed this was never the case. When called in this way,
`Formatter` would use the `result` variable without defining it.

This patch makes the following changes:

- Add a step 4 to the `Formatter` algorithm to explicitly handle a lack
of format specifier.

- Correct the indentation of what was step 3.8 of `Formatter`, so it is
now step 6.

- Remove what was step 4 from `Formatter`, since this is now redundant.

- Remove step 5 from the `Logger` algorithm, since this is now
redundant.

- Add a new step 1 to `Formatter`, which returns early if only one
argument was provided.

---

   * @terinjokes 
   * @domfarolino 
   * No tests added. This is a correction to the spec, to match what was already intended.

***
<a href="https://whatpr.org/console/205.html" title="Last updated on Dec 14, 2021, 10:32 AM UTC (aa06510)">Preview</a> | <a href="https://whatpr.org/console/205/6447e9d...aa06510.html" title="Last updated on Dec 14, 2021, 10:32 AM UTC (aa06510)">Diff</a>

***
<a href="https://whatpr.org/console/205.html" title="Last updated on Sep 12, 2022, 10:18 AM UTC (a251358)">Preview</a> | <a href="https://whatpr.org/console/205/02a69ee...a251358.html" title="Last updated on Sep 12, 2022, 10:18 AM UTC (a251358)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -331,7 +331,6 @@ more arguments are left. It returns a [=/list=] of objects suitable for printing
      |converted|.
 1. Let |result| be a [=/list=] containing |target| together with the elements of |args| starting
    from the third onward.
-1. If |result|'s [=list/size=] is 1, return |result|.
 1. Return <a abstract-op>Formatter</a>(|result|).
 
 <h4 id="formatting-specifiers">Summary of formatting specifiers</h4>
--- a/index.bs
+++ b/index.bs
@@ -287,8 +287,6 @@ describes how it processes format specifiers while doing so.
 1. Let |rest| be all elements following |first| in |args|.
 1. If |rest| is [=list/is empty|empty=], perform
    <a abstract-op>Printer</a>(|logLevel|, « |first| ») and return.
-1. If |first| does not contain any format specifiers, perform
-   <a abstract-op>Printer</a>(|logLevel|, |args|).
 1. Otherwise, perform <a abstract-op>Printer</a>(|logLevel|, <a abstract-op>Formatter</a>(|args|)).
 1. Return *undefined*.
 
@@ -308,9 +306,12 @@ The formatter operation tries to format the first argument provided, using the o
 will try to format the input until no formatting specifiers are left in the first argument, or no
 more arguments are left. It returns a [=/list=] of objects suitable for printing.
 
+1. If |args|'s [=list/size=] is 1, return |args|.
 1. Let |target| be the first element of |args|.
 1. Let |current| be the second element of |args|.
 1. Find the first possible format specifier |specifier|, from the left to the right in |target|.
+1. If no format specifier was found, return |args|.
+1. Otherwise:
   1. If |specifier| is `%s`, let |converted| be the result of
      [$Call$](<a idl>%String%</a>, **undefined**, « |current| »).
   1. If |specifier| is `%d` or `%i`:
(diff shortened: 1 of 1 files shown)


AtkinsSJ:
Realised another small change I forgot to make, will update this when I'm back at my desk.

AtkinsSJ:
New changes: Remove a redundant step from `Formatter`, and make the commit message clearer.

AtkinsSJ:
Just realised in the shower: Formatter also doesn't handle the case of being called with only 1 arg, which again happens when called from `trace()`, `group()` and `groupCollapsed()`.

Probably a better solution is to forget the changes I've made here, and instead make those 3 places call `Logger`, since that already handled these cases.

foolip:
@domfarolino @robertkowalski @terinjokes can any of you take a look? (@AtkinsSJ just asked on the Matrix channel.)

terinjokes:
I can take a look this evening.

AtkinsSJ:
@terinjokes Any thoughts on this so far?

terinjokes:
LGTM. @domfarolino ?

randydwhitephotography (commented):
7457

randydwhitephotography (commented):
KU7IK,H\J
  
GH'
'GH
GH
'''''HGHGH

domfarolino:
> LGTM. @domfarolino ?

Yep I think this LGTM as well, but I'll defer to you @terinjokes since you got to this first!

AtkinsSJ:
Just remembered this PR again, and rebased it on the main branch. Is there anything I still need to do to get verified as a WHATWG participant? I registered for that quite a while ago.

domfarolino:
Looks like you already [signed](https://github.com/whatwg/participant-data/blob/main/individuals.json#L3547-L3557) the participant agreement and I've gone ahead and [verified](https://github.com/whatwg/participant-data/commit/173961c5c3f395d0fc01830e03419859c7438dcb) you as well. Looks like the build is acting up so let me close this and re-open it to see if that helps.

index.bs
```diff
@@ -328,9 +329,8 @@ more arguments are left. It returns a [=/list=] of objects suitable for printing
   1. <p class="XXX">TODO: process %c</p>
   1. If any of the previous steps set |converted|, replace |specifier| in |target| with
      |converted|.
-  1. Let |result| be a [=/list=] containing |target| together with the elements of |args| starting
-     from the third onward.
-1. If |target| does not have any format specifiers left, return |result|.
+1. Let |result| be a [=/list=] containing |target| together with the elements of |args| starting
+   from the third onward.
 1. If |result|'s [=list/size=] is 1, return |result|.
```
domfarolino, line 334: Can we get rid of this line "1. If |result|'s [=list/size=] is 1, return |result|." now that we're handling the args.length == 1 case as the first step of Formatter?

AtkinsSJ: Thanks for looking at this. You're right, that's redundant now. I've removed it.

domfarolino (approved):

domfarolino:
Thanks a lot @AtkinsSJ!

AtkinsSJ:
Thanks for the merge! :^)

PR #210, opened 2022-01-24
Meta: update repository files

annevk:
See https://github.com/whatwg/spec-factory for details.


domfarolino (approved):

domfarolino:
Sorry I just came across this. @annevk do you know what is happening here with the error?

annevk:
It looks like there is an `img` without `width` and `height` attributes.

domfarolino:
At least in the spec src, that seems to be consistent with other `` tag usages of the other images: https://github.com/whatwg/console/blob/main/index.bs#L486.

I have no idea what this error means, since the image exists and is referenced just like all of the other images.

domfarolino:
I tried:
 - Replacing the image locally with a new one and running `make remote`: Still got the dimensions error
 - Deleting the image (and references to it in `index.bs`): Got the dimensions error for a _different_ error in the `images/` directory

This makes me think that something is broken with these "extra files" and Bikeshed specifically when it is used in a remote fashion. So I ran `make local` and everything worked.

Maybe something with bikeshed broke recently, specifically when handling extra files remotely. @tabatkins can you take a look perhaps?

annevk:
It got stricter about images. So we either have to make the images available to the build environment or we add `width`/`height` attributes to all of them. The latter is how we fixed a similar issue elsewhere and how I suggest we proceed here.

cc @domenic as fyi

tabatkins:
Yes, it's a Good Thing to provide image dimensions in the source (prevents layout shift as the images load), and Bikeshed will automatically compute and insert those *if* it can find the image locally; if it can't, it'll complain at you to insert them yourself. This was added last year, I believe, so specs which haven't been touched in a while might not be clean in this regard.

PR #218, opened 2022-12-06
Add note about escaping `%`

tomayac:
It looks like a bit of an interop issue:

Chrome:

Screenshot 2022-12-06 at 12 01 42

Safari:

Screenshot 2022-12-06 at 12 01 15

Firefox:

Screenshot 2022-12-06 at 12 01 26

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -375,6 +375,11 @@ The following is an informative summary of the format specifiers processed by th
     <td>Applies provided CSS</td>
     <td>n/a</td>
   </tr>
+  <tr>
+    <td>`%%`</td>
+    <td>Escapes the `%` specifier</td>
+    <td>n/a</td>
+  </tr>
 </table>
 
 <h3 id="printer" abstract-op lt="Printer">Printer(|logLevel|, |args|[, |options|])</h3>


terinjokes:
This makes sense to me, but curious if there's comments from Firefox or Safari.

photopea:
I have been using the web console to "print an object" and copy-paste it from a web console into a file since 2010 (for debugging, comparing two versions of data), and it is suddenly broken.

    console.log(JSON.stringify( {a:"%%Hi",b:"%Hey"} ));
    =>  {"a":"%Hi","b":"%Hey"}

nchevobbe:
If I'm reading the spec:

> If args’s [size](https://infra.spec.whatwg.org/#list-size) is 1, return args.

So the screenshot in https://github.com/whatwg/console/pull/218#issue-1477863265 for Chrome seems to be wrong for me. Did Chrome behavior changed recently ?

That being said, if you test with multiple parameters, as expected when using specifiers `console.log("%%s", "hello")`, Chrome, Firefox and Safari do print `%s hello`.

So, the change here makes sense to me, but I guess a bug should be filed on Chrome for `console.log("%%")`

HolgerJeromin:
I found a similar issue:
```ts
console.error( 'message: %s', '%hello%%world%');
```
Chrome:
> message: %hello%world%

Firefox:
> message: %hello%%world%

domfarolino:
I think so, yes. I guess this PR just needs the appropriate changes to `Formatter()` (beyond the informative table entry)? Also @terinjokes's [comment](https://github.com/whatwg/console/pull/218#issuecomment-1339479193) is interesting about `Formatter()` not even being invoked for the case in the OP at all, which seems important.

domfarolino:
I'd appreciate a first stab at it if possible, as I probably won't have time myself to drive the changes for a little while. But the algorithms you'd need to touch aren't too complicated I promise!

photopea:
console.log() with a single parameter should print exactly the string which it received. 

When I have an object in JS, I want to save it into a JSON file. I do

     console.log(JSON.stringify(str)); 

And when I copy-paste it from a console to a text file, I get a different object.

tomayac:
So looking back at the original post https://github.com/whatwg/console/pull/218#issue-1477863265, which of the two behaviors is considered correct?

terinjokes:
As noted in https://github.com/whatwg/console/pull/218#issuecomment-1339479193, the current specification is to _not_ intepret `%` if there are no rest parameters. IIRC, this case was put into `Logger` to support logging arbitrary strings in a backwards compatible manner.

Issue #148, opened 2018-08-05
Format specifier for BigInt


shisama:
Chrome 68 supports BigInt and also Node.js v10 supports it with `--harmony-bigint` flag. But console.log doesn't support BigInt in them.

Chrome 68

```js
> console.log("%d", 1180591620717411303424n)
NaN
> console.log("%i", 1180591620717411303424n)
NaN
```

Node.js v10.7.0 with `--harmony-bigint`

```js
> console.log("%d", 1180591620717411303424n)
1.1805916207174113e+21
> console.log("%i", 1180591620717411303424n)
1.1805916207174113e+21
```

I think expected result is 

```js
> console.log("%d", 1180591620717411303424n)
1180591620717411303424n
> console.log("%i", 1180591620717411303424n)
1180591620717411303424n
```

However, [specs](https://console.spec.whatwg.org/#formatting-specifiers) shows `%d or %i` uses `%parseInt%(element, 10)` for type conversion.
What specifier is the best for BigInt? Has this already been discussing?

silverwind:
Keep in mind that Node.js is currently violating the spec for the `%d` specifier where it uses the `Number` constructor. The specced `parseInt` on the otherhand can output numbers without scientific notation because its [toString](https://tc39.github.io/proposal-bigint/#sec-tostring-applied-to-the-bigint-type) method which does not include the notion of a scientific notation.

So I think, we may not even need a spec change.

terinjokes:
I believe the issue isn't `NumberToString` (since `BigInt` isn't a Number), but `parseInt`'s call to `ToInt32`.

```
%parseInt%(1180591620717411303424n, "10") <- 1.1805916207174113e21
  ToString(1180591620717411303424n) <- "1180591620717411303424" (https://tc39.github.io/proposal-bigint/#sec-tostring-applied-to-the-bigint-type)
  ToInt32("1180591620717411303424") <- 1.1805916207174113e21
   ToNumber("1180591620717411303424") <- 1.1805916207174113e21
```

I would prefer to define `%d` and `%i` specifiers to handle `BigInt`. The current description does not state int32, but instead "Element which substitutes is converted to an integer" which is vague enough.

On the other hand, `BigInt.ToNumber()` is defined to throw an error, and I'm unsure if using the same formatters will cause confusion with developers.

silverwind:
> Also the third column of https://console.spec.whatwg.org/#formatting-specifiers should be updated

[...]

PR #224, opened 2023-07-12
Handle BigInt with the `%d` and `%i` format specifiers

domfarolino:
This PR aims to close https://github.com/whatwg/console/issues/148 by including special-handling for `BigInt` in the `%d` and `%i` format specifier logic, that defers to the ECMAScript `BigInt::toString()` abstract operation, which provides a stringified representation of the BigInt that's optimal for that type.

 - Firefox already implements this behavior exactly
 - Safari _almost_ implements this; they add fancy commas that the `BigInt::toString()` abstract operation doesn't add
 - Node.js also _almost_ implements this PR, but adds the `n` suffix to the logged output, which this PR does not call for

All of those implementations specifically do not implement what is standardized right now, which is calling `%parseInt()%` on BigInts formatted with the `%d` and `%i` specifier, so the shortest path to interoperability is likely to change the spec here (and nudge other implementations to slightly modify their behavior to match this PR, which they're so close to doing). The only browser that implements the _current spec_ behavior is Chromium, and I personally like what the other implementations are doing better. The ecosystem has seemed to naturally drift in the the direction of the non-Chromium implementation too.

(The below template will be filled out provided this PR is acceptable).

   * Firefox (already implements this behavior)
   * Safari is quite close. We can see if they'd be willing to change to strictly adhere to this spec
   * Node.js is quite close. We can see if they'd be willing to change to strictly adhere to this spec
   * Need manual tests
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: …
   * Node.js: …

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -17,6 +17,8 @@ urlPrefix: https://tc39.github.io/ecma262/#; spec: ECMASCRIPT
     text: Type; url: sec-ecmascript-data-types-and-values
     text: ToString; url: sec-tostring
     text: Call; url: sec-call
+    for: BigInt
+      text: toString; url: sec-numeric-types-bigint-tostring
   type: interface
     text: %ObjectPrototype%; url: sec-properties-of-the-object-prototype-object
   type: constructor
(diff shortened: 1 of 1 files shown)


domfarolino:
Hey @terinjokes if you could take a look at this when you get time that'd be great. I opted to spec what was described in https://github.com/whatwg/console/issues/148#issuecomment-411851902, which conveniently seems to be almost exactly what most implementations are doing. This is not the direction that you mentioned elsewhere in https://github.com/whatwg/console/issues/148 about adding a new format specifier, but I personally prefer this path forward and would like your feedback on this direction.

dcrousso:
i think WebKit shows a comma because it uses `Number.prototype.toLocaleString` for `%i`/`%d`/`%f`

seems reasonable to just remove that (not to mention `console.log("1000")` should probably have the same output as `console.log("%i", 1000)`)

domfarolino:
Oh yeah I didn't realize Safari was doing that unconditionally, even for non-BigInts. I agree it seems reasonable to remove that. I've filed https://bugs.webkit.org/show_bug.cgi?id=259146 independently.

Issue #140, opened 2018-05-26
Spec console.timeStamp


domfarolino:
See https://github.com/whatwg/console/issues/27.

and-oli:
Reusing this open issue to discuss a proposal to explore potential enhancements to the console.timeStamp API, which is a non-standard but it’s implemented in several major browsers. The goal is to improve its utility for customizing the Performance panel in developer tools, particularly in the context of modern web frameworks that often abstract the browser's underlying model. (This proposal is being submitted on behalf of the Chrome Page Quality team at Google).
## Explored Changes:
**Custom Measurements:** Allow developers to add custom measurements to the Performance panel using console.timeStamp, in addition to the current instant markers. This could involve additional arguments to the API, such as a measurement name and a way to indicate the start and end time of the measurement.
**Swimlane Control:** Enable developers to specify the swimlane in which their custom measurements appear. This could be achieved through another argument to the API or a separate configuration option.
**Custom presentation format:** A generalization of the point above, so that other presentation parameters like entry color, details, etc. can be customized as well.

## Use Cases:

Debugging Complex Apps: In large applications, custom measurements could help isolate performance bottlenecks related to specific components or interactions.

Framework and tools Integration: Developers using frameworks could add framework-specific performance metrics to the Performance panel, aligning it with their mental models.

## Considerations:
* Standardization: As console.timeStamp is not currently standardized, any changes should be considered for potential future standardization.
* Compatibility: The proposed enhancements should aim to be compatible with existing implementations in major browsers, if possible.

## Alternative Approaches and why `console.timeStamp`:
We considered extending the existing performance.mark and performance.measure APIs. However, these APIs are designed to create user space-observable data (the data is buffered and readable / observable in the page), and maintaining this infrastructure introduces a slight overhead, which is particularly noticeable when the API is called extensively. Since the sole purpose of console.timeStamp is to add data to the DevTools Performance panel, it can be optimized for minimal overhead, particularly in cases where profiling is disabled. Additionally this functionality, being primarily for interaction with DevTools, aligns better with the console API.

## Potential API Prototype:
###  Overloading console.timeStamp
```ts
type TimeStampName = string;
// Current impl: Single instant marker.
console.timeStamp(timeStampName: TimeStampName);

// Override: Measurement with custom start and options.
console.timeStamp(
  measureName: TimeStampName,
  start: TimeStampName,
  devtoolsOptions?: DevToolsOptions
);

// Example
// t = 0
console.timeStamp("start");
[...]

Issue #27, opened 2016-01-10
Missing methods: dir, dirxml, markTimeline, timeStamp, timeline, timelineEnd, exception


domenic:
These seem to exist in a few different browsers.

terinjokes:
What benefit do `dir` and `dirxml` have over format specifiers?

@paulirish had mentioned at one point that `markTimeline`, `timeStamp`, `timeline`, and `timelineEnd` (and simpler) were going to be removed from Chrome, and with the uncertainty I didn't include them. This decision seems to have been reversed, as they still exist.

I haven't previously considered `exception`.

domenic:
I doubt they have any benefit, but they're implemented, so we gotta spec 'em.

exception is not in Chrome but is in Edge and Firefox.

paulirish:
`timeline` and `timelineEnd` are deprecated and nonfunctional in Chrome. they were unique to us. i would exclude from the spec.

`markTimeline` is deprecated, but functional. it's equivalent to `timeStamp`. this was also unique to chrome.

`timestamp` we do use, and [Edge does](https://dev.windows.com/en-us/microsoft-edge/platform/documentation/f12-devtools-guide/console/using-the-console-api/) as well.

domenic:
Hmm Firefox supports timeline and timelineEnd. I wonder if we'll need to spec them as no-ops to avoid crashy pages.

markTimeline is also in Firefox.

Found http://mxr.mozilla.org/mozilla-central/source/dom/webidl/Console.webidl ... it looks like markTimeline, timeline, and timelineEnd are specifically no-ops there.

paulirish:
`dir` and `dirxml` are useful when logging DOM/SVG/XML elements.  dir will log the object representation (like a JS object). dirxml will log a tree like what you see in the DOM/Elements panel.

terinjokes:
@paulirish and we have formatters in the spec for both behaviors already.

<blockquote>
<h3>3.2. Formatting specifiers that require a complex UI</h3>
<table>
  <thead>
    <tr>
     <th>Specifier</th>
     <th>Function</th>
     <th>Type Conversion</th>
  </thead>
  <tbody>
   <tr>
     <td>%o</td><td>Displays as expandable DOM element</td><td>n/a</td>
   </tr>
   <tr>
     <td>%O</td><td>Displays as expandable JS Object</td><td>n/a</td>
   </tr>
  </tbody>
</table>
</blockquote>

domenic:
[...]

PR #236, opened 2024-07-12
Add `console.timeStamp()` to the specification

bmeurer:
This adds a minimal (vague) definition for the `timeStamp()` method, based on its current behavior across Chromium, Firefox, and Safari, and the MDN documentation[^1].

[^1]: https://developer.mozilla.org/en-US/docs/Web/API/console/timestamp_static

Ref: #140

- [x] At least two implementers are interested (and none opposed):
   * Google
   * Mozilla
- [x] Implementations already exist:
   * Chromium
   * Geck
   * WebKit
- [x] [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/console/timestamp_static) already exists.
- [x] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -66,6 +66,7 @@ namespace console { // but see namespace object requirements below
   undefined time(optional DOMString label = "default");
   undefined timeLog(optional DOMString label = "default", any... data);
   undefined timeEnd(optional DOMString label = "default");
+  undefined timeStamp(optional DOMString label = "default");
 };
 </pre>
 
@@ -273,6 +274,12 @@ Each {{console}} namespace object has an associated <dfn>timer table</dfn>, whic
 for plans to make {{console/timeEnd()}} and {{console/timeLog()}} formally report warnings to the
 console when a given |label| does not exist in the associated <a>timer table</a>.
 </p>
+
+<h4 id="timestamp" oldids="timestamp-label,dom-console-timestamp" method for="console">timeStamp(|label|)</h4>
+
+1. If the developer is recording a performance trace, add a single marker to the performance trace with the |label|.
+1. Otherwise, do nothing.
+1. Return *undefined*.
 
 <h2 id="supporting-ops">Supporting abstract operations</h2>
 


bmeurer:
@domfarolino Let's get the ball rolling here. :muscle: 

@and-oli Please loop in your contacts with other browser vendors here.

domfarolino:
Hmm, it looks like you removed [this line about tests](https://github.com/whatwg/console/blob/main/PULL_REQUEST_TEMPLATE.md?plain=1#L10-L11) from the pull request template? Can you add that back and maybe write some tests for this API? You can follow the pattern in https://github.com/web-platform-tests/wpt/tree/master/console. Note the state of testing is a little sad at this point, since these are basically just manual tests :(

index.bs
```diff
@@ -274,6 +275,12 @@ for plans to make {{console/timeEnd()}} and {{console/timeLog()}} formally repor
 console when a given |label| does not exist in the associated <a>timer table</a>.
 </p>
 
+<h4 id="timestamp" oldids="timestamp-label,dom-console-timestamp" method for="console">timeStamp(|label|)</h4>
```
domfarolino, line 278: Maybe this should go in its own section about performance APIs? I guess this is kind of time related, but [the existing timing APIs](https://console.spec.whatwg.org/#timing) all have to do with keeping track of literal time stamps that get printed for the developer, where as `timeStamp()` seems to not interact with that timer table at all, and be more focused on interacting with a performance console.

This sounds very similar [to `console.markTimeline()`](https://github.com/whatwg/console/issues/27#issuecomment-171005496) actually. Am I correct in thinking it does not deal with actual raw times or timestamps at all?

(Also this is separate, and maybe I am dumb, but can you show me where the marked point actually shows up in the Performance panel's UI, when this API is called? I can't seem to find it when I call the API)

and-oli:
@dcrousso @nchevobbe

index.bs
```diff
@@ -66,6 +66,7 @@ namespace console { // but see namespace object requirements below
   undefined time(optional DOMString label = "default");
   undefined timeLog(optional DOMString label = "default", any... data);
   undefined timeEnd(optional DOMString label = "default");
+  undefined timeStamp(optional DOMString label = "default");
```
dcrousso, line 69: im not sure exactly the behavior of Chrome/Firefox/etc., but at least in WebKit there is no `= "default"` value (i.e. calling `console.timeStamp()` will show "Timestamp" in the UI)

index.bs
```diff
@@ -274,6 +275,12 @@ for plans to make {{console/timeEnd()}} and {{console/timeLog()}} formally repor
 console when a given |label| does not exist in the associated <a>timer table</a>.
 </p>
 
+<h4 id="timestamp" oldids="timestamp-label,dom-console-timestamp" method for="console">timeStamp(|label|)</h4>
+
+1. If the developer is recording a performance trace, add a single marker to the performance trace with the |label|.
```
dcrousso, line 280: im slightly concerned that this is the first time we've mentioned anything like "performance trace" in this spec.  this feels a bit limiting to me (e.g. what if instead of showing a marker we'd like to put it in a table?  how does this work with non-GUI environments?  etc.).  is there a way we can make this a bit more generic about just "capture a timestamp (relative to XYZ) for use in an implementation-defined way"?

index.bs
and-oli: > can you show me where the marked point actually shows up in the Performance panel's UI, when this API is called?

In Chrome, any call to console.timeStamp that happens during a recording is shown inside the "Timings" track.
image

and-oli: and yeah at least in Chrome, you are right about `timeStamp` being independent from the other timing APIs in `console`. There, it simply adds an entry to the data buffer of the performance panel (a trace event)

domfarolino: Got it, can you confirm whether or not other browsers behave the same (or at least are as distinct from the other timing APIs as in Chrome)? That will help inform the tests we'll need for this I think.

index.bs
domfarolino: Yeah, Chromium appears to not use "default" at all. At least when provided with no string, it shows `"TimeStamp: <empty string technically here>"` in the UI which I think matches Safari, but I fear I am too dumb to figure out where in the "Timelines" UI calls to `console.timeStamp()` actually show up.

Screenshot 2024-07-21 at 11 09 01 PM

index.bs
domfarolino: Hmm, at first I strongly agreed, but when I looked into what [Node](https://nodejs.org/api/console.html#consoletimestamplabel) and [Deno](https://docs.deno.com/api/web/~/Console.timeStamp) do, they specifically mention the method is a noop unless hooked up to an inspector environment with a timeline UI. Maybe this isn't too bad to codify directly in the spec, since it matches reality with every single implementation, including non-browser ones.

But if people feel strongly, we can avoid that, making the normative text more vague, and adding some non-normative text mirroring what common developer documentation mentions about this API — namely that in browser/UI environments, it generally targets a performance-ey panel of sorts.

index.bs
bmeurer: Yes, I just fixed V8 last week to have the `= "default"` value for all `console.time*()` APIs, including `console.timeStamp()`. Maybe we can update JSC as well to have consistency.

index.bs
bmeurer: According to MDN, `console.timeStamp()` is solely intended for the purpose of adding data to performance traces, and from my tests with Firefox, that's exactly what happens there.

@and-oli can you confirm for Safari?

index.bs
bmeurer: We can also make this more vague, but I thought it'd be useful to capture reasonable expectations for the API in the spec, so developers know that for conforming implementations

1. this is a no-op when no trace is being recorded, and
2. when a trace is being recorded, calls to `console.timeStamp()` will leave a mark in the trace in some form.

index.bs
```diff
@@ -274,6 +275,12 @@ for plans to make {{console/timeEnd()}} and {{console/timeLog()}} formally repor
...
+1. Otherwise, do nothing.
```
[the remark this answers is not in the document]
bmeurer: Ack.

index.bs
```diff
@@ -274,6 +275,12 @@ for plans to make {{console/timeEnd()}} and {{console/timeLog()}} formally repor
...
+1. Return *undefined*.
```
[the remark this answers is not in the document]
bmeurer: Ack.

index.bs
bmeurer: @domfarolino Let me know if you want me to put this into a separate section on Performance APIs.

bmeurer:
> Hmm, it looks like you removed [this line about tests](https://github.com/whatwg/console/blob/main/PULL_REQUEST_TEMPLATE.md?plain=1#L10-L11) from the pull request template? Can you add that back and maybe write some tests for this API? You can follow the pattern in https://github.com/web-platform-tests/wpt/tree/master/console. Note the state of testing is a little sad at this point, since these are basically just manual tests :(

Beyond testing that the function exists on the `console` object, how would we go about adding other (meaningful) tests?

index.bs
dcrousso: i kinda prefer it showing "Timestamp" in the UI, but i dont feel too strongly about it

we could always have the API use `= "default"` and then in the UI check if `=== "default"` and if so show "Timestamp"

bmeurer: Ok, so we can stick to `= "default"` for consistency here?

index.bs
bmeurer: Friendly ping.

bmeurer:
> > Hmm, it looks like you removed [this line about tests](https://github.com/whatwg/console/blob/main/PULL_REQUEST_TEMPLATE.md?plain=1#L10-L11) from the pull request template? Can you add that back and maybe write some tests for this API? You can follow the pattern in https://github.com/web-platform-tests/wpt/tree/master/console. Note the state of testing is a little sad at this point, since these are basically just manual tests :(
> 
> Beyond testing that the function exists on the `console` object, how would we go about adding other (meaningful) tests?

@domfarolino any thoughts here?

index.bs
and-oli: Sorry for the delay. Safari also uses console.timeStamp solely to add markers to the Timelines tab: [see docs](https://webkit.org/web-inspector/console-object-api/#:~:text=console.timeStamp(%5Bmessage%5D)%20adds%20a%20marker%20to%20the%20active%20timeline%20recording.)

index.bs
domfarolino: > we could always have the API use = "default" and then in the UI check if === "default" and if so show "Timestamp"

This seems a little funky, especially since Safari treats `console.timeStamp()` differently from `console.timeStamp("default")`—the former defaults to "TimeStamp" and the latter prints "default". Whereas Chrome shows "TimeStamp: default" for both. I think it makes more sense to stick with `= "default"` personally. Do you think Safari's inspector would be willing to budge on this @dcrousso ?

index.bs
domfarolino: Yeah, I think I'm OK with what's currently spec'd here since it matches reality. We can always expand if needed, i.e., if non-GUI environments start to do something special here. I'll resolve this but feel free to re-open if you feel strongly.

domfarolino:
Sadly I don't think we can do much more than add a few manual tests to WPT, and do whatever vendor-specific testing you can do. I think this PR is good to go % https://github.com/whatwg/console/pull/236/files#r1676151812. @bmeurer does Firefox match Chrome in this case? (I couldn't find where Firefox puts timeStamp() marks on its performance timeline). If it does match Chrome, we can just land this since it would match 2/3 implementations without strong objections from another.

index.bs
dcrousso: like i mentioned, im not a huge fan of showing "default" in the UI as i think it's not super clear (especially when compared to showing "Timestamp"), but Web Inspector certainly can adopt the behavior of `DOMString label = "default"` and then have the UI adjust based on whether a value was actually provided

tunetheweb:
> @bmeurer does Firefox match Chrome in this case? (I couldn't find where Firefox puts timeStamp() marks on its performance timeline). If it does match Chrome, we can just land this since it would match 2/3 implementations without strong objections from another.

I've confirmed that Firefox removed UI support for this when they launched their new profiler but they used to support it. They still accept the API but currently don't use that information. Bug to add support here: https://bugzil.la/1730896

MDN updated here (with support from @canova from Firefox Profiler team): https://github.com/mdn/browser-compat-data/pull/28074

Not sure what that "half-support" in Firefox, and the small outstanding "default" issue in Safari means in terms of "2/3 implementations without strong objections from another" though? I'd argue it meets the bar and with some support in all three browsers it would be good to merge this and add it to the spec.

PR #244, opened 2025-05-26
Specify console.context

### LINKED ISSUE #193: console.context()
In the spirit of https://github.com/whatwg/console/issues/27 and https://github.com/whatwg/console/issues/74 I'm filing issues about two additional console members I've spotted as part of https://github.com/foolip/mdn-bcd-collector/issues/917.

`console.context(name)` is available in Chrome and Node.js, and returns a new object with a lot of the `console` members. This test gives some clues about what it does:
https://chromium.googlesource.com/v8/v8/+/23d0a6a5125d407655f2b7f6bb7263b4e05019d8/test/inspector/runtime/console-context.js

It looks like `console.context('bla').log('hello')` doesn't log anything in Node.js, but logs "hello" just like `console.log("hello")` would in Chrome.

Does anyone recognize what this is, and would it be useful to standardize? I expect not, but want to check.

nchevobbe:
This specifies the console.context(label) method which returns a new console namespace object with an optional context name label.

Tests: TODO

Fixes #193

---

- [x] At least two implementers are interested (and none opposed):
   * Chrome is shipping the method already 
   * Mozilla is interested
   * TODO (waiting to get some feedback on the PR before looking into writing tests)
- [x] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: https://issues.chromium.org/issues/40523727
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1948870
   * Node.js: https://github.com/nodejs/node/issues/56634 (Node.js does have support for `console.context()` but logging with returned instance doesn't print anything. Linked issue mentions the lack of spec for the method).
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=293768
   * Deno:https://github.com/denoland/deno/issues/29560
- [x] MDN issue: https://github.com/mdn/mdn/issues/692
- [x] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use.

***
<a href="https://whatpr.org/console/244.html" title="Last updated on Nov 17, 2025, 8:35 AM UTC (5da9ceb)">Preview</a> | <a href="https://whatpr.org/console/244/6f299db...5da9ceb.html" title="Last updated on Nov 17, 2025, 8:35 AM UTC (5da9ceb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -302,8 +302,8 @@ Each {{console}} namespace object has an associated <dfn>context name</dfn>, emp
 
 1. Let |context| be a new {{console}} namespace object.
 1. Let |contextName| be the empty string.
-1. If |label| is not null, let |contextName| be |label|.
-1. Let |context|'s <a>context name</a> be |contextName|.
+1. If |label| is not null, then set |contextName| to |label|.
+1. Set |context|'s <a>context name</a> to |contextName|.
 1. Return |context|.
 
 <h2 id="supporting-ops">Supporting abstract operations</h2>
--- a/index.bs
+++ b/index.bs
@@ -83,6 +83,9 @@ namespace console { // but see namespace object requirements below
   undefined time(optional DOMString label = "default");
   undefined timeLog(optional DOMString label = "default", any... data);
   undefined timeEnd(optional DOMString label = "default");
+
+  // Contextualizing
+  console context(optional DOMString label = "");
 };
 </pre>
 
@@ -290,6 +293,18 @@ Each {{console}} namespace object has an associated <dfn>timer table</dfn>, whic
 for plans to make {{console/timeEnd()}} and {{console/timeLog()}} formally report warnings to the
 console when a given |label| does not exist in the associated <a>timer table</a>.
 </p>
+
+<h3 id="contextualizing">Contextualizing</h3>
+
+Each {{console}} namespace object has an associated <dfn>context name</dfn>, empty by default.
+
+<h4 id="context" method for="console">context(|label|)</h4>
+
+1. Let |context| be a new {{console}} namespace object.
+1. Let |contextName| be the empty string.
+1. If |label| is not null, let |contextName| be |label|.
+1. Let |context|'s <a>context name</a> be |contextName|.
+1. Return |context|.
 
 <h2 id="supporting-ops">Supporting abstract operations</h2>
 
@@ -512,6 +527,9 @@ enhancements:
       <img alt="Indicating UI that allows filtering by log severity" width="494" height="58"
       src="images/severity-filter.png">
     </div>
+  </li>
+  <li>
+    Extra UI allowing the user to filter messages by context name
   </li>
   <li>
     Extra UI off to the side indicating the current state of the <a>timer table</a>,
--- a/index.bs
+++ b/index.bs
@@ -85,7 +85,7 @@ namespace console { // but see namespace object requirements below
   undefined timeEnd(optional DOMString label = "default");
 
   // Contextualizing
-  console context(optional DOMString label = "");
+  ConsoleContext context(optional DOMString label = "");
 };
 </pre>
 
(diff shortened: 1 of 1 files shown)


nchevobbe (commented):
Note that this is the first time I'm writing such spec PR and I could be doing things the wrong way.

index.bs
```diff
@@ -291,6 +294,18 @@ for plans to make {{console/timeEnd()}} and {{console/timeLog()}} formally repor
...
 
+<h3 id="contextualizing">Contextualizing</h3>
+
+Each {{console}} namespace object has an associated <dfn>context name</dfn>, empty by default.
+
+<h4 id="context" method for="console">context(|label|)</h4>
+
+1. Let |context| be a new {{console}} namespace object.
+1. Let |contextName| be the empty string.
+1. If |label| is not null, let |contextName| be |label|.
```
annevk, line 305:
```suggestion
1. If |label| is not null, then set |contextName| to |label|.
```

index.bs
```diff
@@ -291,6 +294,18 @@ for plans to make {{console/timeEnd()}} and {{console/timeLog()}} formally repor
...
+1. Let |context|'s <a>context name</a> be |contextName|.
```
annevk, line 306:
```suggestion
1. Set |context|'s <a>context name</a> to |contextName|.
```

annevk (commented):
These are drive-by comments. I don't have enough context to do a full review.

domfarolino:
> So I guess it should return some kind of Console/Context/ConsoleInstance interface, that hopefully we can then reuse for the definition of the console namespace?

What does Chromium do? Maybe @bmeurer knows?

domfarolino:
I'm happy to keep the design discussion in https://github.com/whatwg/console/issues/193, since that's where most of it has been so far. Just wanted to point out that given https://github.com/whatwg/console/issues/193#issuecomment-2952919359, I think we need to think harder about how to specify the interaction between internal state like the group stack and the notion of multiple console objects (whether they come from different documents, or the same document via `.context()`).

bmeurer:
> > So I guess it should return some kind of Console/Context/ConsoleInstance interface, that hopefully we can then reuse for the definition of the console namespace?
> 
> What does Chromium do? Maybe @bmeurer knows?

That's basically what Chromium does, see [`builtins-context.cc`](https://source.chromium.org/chromium/chromium/src/+/main:v8/src/builtins/builtins-console.cc;l=264-316;drc=5e74ac5d42bce00b534d992272e7fa09b53b300f).

nchevobbe:
I did something very simple here, defining a `ConsoleContext` interface with its own webidl, linking the methods to the one already defined for the `console` namespace object.

Let me know if there's something better we can do, for example, i'm not sure it's okay that the methods definition for both `console` and `ConsoleContext` are still under the `console` "section".

---

> I'm happy to keep the design discussion in #193, since that's where most of it has been so far. Just wanted to point out that given [#193 (comment)](https://github.com/whatwg/console/issues/193#issuecomment-2952919359), I think we need to think harder about how to specify the interaction between internal state like the group stack and the notion of multiple console objects (whether they come from different documents, or the same document via `.context()`).

For now, I'm aiming at specifying what's shipping in Chrome in the sake of moving this forward, and we can have discussions in specific issues regarding the questions that this brings (I think that's okay, `context()` doesn't add entirely new problems, we already have those with multiple documents)

nchevobbe:
Let's see if we can get this moving forward, pinging a few DevTools folks so we can have views from different vendors @oslego  @danilsomsikov @leahmsft

danilsomsikov:
Hi @nchevobbe, thanks for the ping. From the Chromium perspective, I don't have any objections to the current direction of the PR.

To address your specific points:

* Interface vs. Namespace: Moving to a ConsoleContext interface is a good call and aligns with how we’ve approached the implementation in Chromium. It feels more robust than a generic namespace object for this purpose.
* Spec Organization: Keeping the method definitions under the existing console section seems like a pragmatic way to avoid duplication for now. As long as the WebIDL linkage is clear, I don't see it as a major issue.
* State Interaction: I agree that the interaction with internal states like the group stack needs careful thought, but I don't think it should be a blocker for the initial specification of .context(). Since similar challenges already exist with multiple documents, we can iterate on this in a follow-up issue

